PR #1, opened 2016-11-10
Fix for , in Window title

cdloh:
If you have over 999 unread email you get a , in your unread count. ie 1,001

Updated regex to match.

Not sure if you get more commas the more unread you have.


timche:
Cool, thanks! :)

PR #3, opened 2016-11-11
add mailto support

cdloh:
Add mailto support to the app.

I've added a new menu to the app called settings. Ideally it would only need one button and track if MailTo support has been enabled already yet.

Would you want that change for the button or are you happy as is?

Edit: Made the above menu change.


timche:
Cool! I'll check it out and merge asap :) 👍

PR #20, opened 2019-02-16
Persist window size

markypython:


src/state/window.js
```diff
@@ -0,0 +1,71 @@
...
+  width: 1280,
+  height: 800
+}
+
+class WindowState {
+  constructor(stateName, window) {
+    this.stateName = `state.window.${stateName}`
+    this.window = window
+
+    console.log(appConfig.get(this.stateName))
```
timche: Leftover?

markypython: Yes.  My mistake

markypython:
@timche As a note, this same concept could be applied to future aspects of the app that require persistence.  We could even create `State` class that things like `WindowState` extend so they can make use of similar functionality such as the `saveState` method.

PR #24, opened 2019-02-17
Fix unread count

markypython:
Currently, the unread count changes based on which label you are in and if you have an email open, the unread count will never change since the title doesn't reflect the unread count.  This aims to solve this problem using the following process.

1. Add a preload script to the main window so we can execute JS code in Gmail.
1. When changes occur, grab the unread count from the inbox label on the left navigation bar

Changes to the unread count are tracked using one of two ways.

1. A MutationObserver that observes the title.  Even though the title does not always work for getting the unread count, it is a good event to listen to for checking the unread count.
1. An interval of 1 second (configurable through the `INTERVAL` const in preload.js) will check the unread count every second.  I tried adding a MutationObserver to the unread count in the sidebar, but it would only work if I went all the way up to the navigation element and observed all subtree changes which was much more intense than a simple 1 second interval.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -1,5 +1,12 @@
 const path = require('path')
-const { app, shell, BrowserWindow, Menu, Tray } = require('electron')
+const {
+  app,
+  ipcMain: ipc,
+  shell,
+  BrowserWindow,
+  Menu,
+  Tray
+} = require('electron')
 const electronDebug = require('electron-debug')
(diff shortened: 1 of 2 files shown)


timche:
Great fix. I have started this project with a very basic implementation for unread count and haven't looked into it further. Thanks a lot 🙏

PR #37, opened 2019-03-20
Only enable debug features when developer mode is enabled

### LINKED ISSUE #33: Targeted debug mode
It would be good to have debug mode only set for development or for specific users (i.e. you and me).  Perhaps reading from an environment variable?

Thoughts?

markypython:
This is not a finished PR as I still would like to explore reloading the menu when the settings change as well as enabling or disabling `electronDebug`.

Fixes #33

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -7,22 +7,14 @@ const {
   Menu,
   Tray
 } = require('electron')
-const electronDebug = require('electron-debug')
-const appConfig = require('electron-settings')
 const { autoUpdater } = require('electron-updater')
 const { is } = require('electron-util')
 
+// Initialize the debug mode handler when starting the app
+require('./debug').init()
+
 const menu = require('./menu')
 const WindowState = require('./state/window')
-
-// Debug mode should only be enabled for development mode
-//   or when a user chooses developer mode
-if (is.development || appConfig.get('developer-mode')) {
-  electronDebug({
-    enabled: true,
-    showDevTools: false
-  })
-}
 
 if (!is.development) {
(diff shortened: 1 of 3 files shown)


src/app.js
```diff
@@ -8,16 +8,21 @@ const {
...
 
 const menu = require('./menu')
 const WindowState = require('./state/window')
 
-electronDebug({
-  enabled: true,
-  showDevTools: false
-})
+// Debug mode should only be enabled for development mode
+//   or when a user chooses developer mode
```
timche: The two extra whitespaces after `//` seem odd to me, is that intended?

markypython:
> Great work! I'm thinking of calling it debug mode rather than developer mode. What do you think?

My thought was that this setting could do more than just enable debug mode, such as enable the developer menu (which I have in this PR).  But if you think it would be better to simply use this setting for enabling/disabling debug mode, I am fine with that and can update.

src/app.js
markypython: That is something I tend to do when I have a long comment broken onto multiple lines to show that it is part of the first sentence.  Then, you can start another sentence and see the difference.

For example.

```js
// Debug mode should only be enabled for development mode
//   or when a user chooses developer mode
// This is another comment that is not part of the previous sentence
```

timche: Oh, alright. Fine for me 👍  Although I would end sentences with a full stop to indicate that. 🙈 But in the end it's personal preference, so I don't mind :)

markypython:
Updated to add a restart prompt when toggling debug mode.  Also, I did update to debug mode instead of developer mode.

PR #38, opened 2019-03-23
Capture notifications

### LINKED ISSUE #15: Sticky notifications not working
After you change the notification style for the Gmail desktop app to "Alert" (=sticky notification). When you receive a new mail, the notification disappears after a few seconds. I like it to be sticky.

markypython:
@timche I got started on the sticky notifications issue.  Making this a draft PR since it is only partially done, but the basic notifications handling is working.  I am capturing the web notifications and sending them to the Electron app so native notifications can be used.

Let me know your thoughts on the direction of this PR.

Fixes #15

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -13,7 +13,7 @@ const { is } = require('electron-util')
 // Initialize the debug mode handler when starting the app
 require('./debug').init()
 // Initialize the notification handler
-require('./notifications')
+require('./notifications').init()
 
 const menu = require('./menu')
 const WindowState = require('./state/window')
--- a/src/app.js
+++ b/src/app.js
@@ -12,6 +12,8 @@ const { is } = require('electron-util')
 
 // Initialize the debug mode handler when starting the app
 require('./debug').init()
+// Initialize the notification handler
+require('./notifications')
 
 const menu = require('./menu')
 const WindowState = require('./state/window')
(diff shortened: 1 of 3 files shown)


markypython:
@timche So, I am still unable to get the openEmail function to work since that is managed using the onclick method of the notification class.  Since that property get's added after the notification is instantiated and web notifications don't use a show notification, there isn't really a way to open the email using the notification click.  That said, we could use the title of the notification to search the emails in the inbox and if there is only one that matches, then we open that email.  The issue with that is of course the inconsistency when there is only one with that title and when there are multiple and the on click changing as a result.  I still think we should move forward with native notifications even if they only open the application, for the following reasons:

1. The current notifications don't support switching to "sticky" style (the whole thing that brought this issue up)
1. Clicking on the current notifications opens a new BrowserWindow which does not have the conditional logic as the main window of opening non Google links in the browser.  This causes issues when opening links to GitHub for example.

Thoughts on all this?  Sorry it got long :)

markypython:
@timche If we move forward with this pr, then no because there shouldn't be any cases where a new window is opened.  Again, the tradeoff here is losing the onclick for the notification but gaining native notifications and not having a new window pop open when clicking on the notification.

timche:
> there shouldn't be any cases where a new window is opened
That's different for me. I like to click on the notification to open the received email. That's an expected behaviour I know also from other email apps (e.g. Airmail).

markypython:
I'm going to close this PR and take another go at trying to get it to work in all the right ways.

PR #40, opened 2019-03-28
TypeScript

### LINKED ISSUE #26: Convert to TypeScript
@timche Thoughts about converting over to TypeScript?

markypython:
Still a work in progress, but this is mostly finished.  @timche Let me know your thoughts on the direction of this PR.

Fixes #26


timche:
I'll review this asap, promised. I'm a little busy these days, sorry.

markypython:
Not a problem.  It is a big change after all!

tsconfig.json
```diff
@@ -0,0 +1,14 @@
...
+    "module": "CommonJS",
+    "target": "es6",
+    "noImplicitAny": true,
+    "removeComments": true,
+    "preserveConstEnums": true,
+    "outDir": "build",
+    "sourceMap": true
+  },
+  "include": ["src/**/*"],
+  "exclude": ["node_modules", "**/*.spec.ts"]
```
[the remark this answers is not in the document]
markypython: I added this so we would be ready for tests, but I can remove.

markypython:
Great thoughts.  I'm fine to fix points 1, 3, and 4, but I would like to discuss point 2 a bit more.  My motivation for changing to Gulp was to setup the application for future improvements and additions.  The css additions in #42 are an example of where we could utilize gulp to minify the css and when we tackle dark mode, it might be nice to use a css preprocessor like SCSS.  Without Gulp, npm scripts can get out of hand and hard to manage, which is why Gulp is a nice option.

Additionally, I am not too concerned about Gulp build tools being out of date for the following reasons:

1. Using features that are so new that build tools haven't implemented them yet is probably a bit too bleeding edge
1. VS Code uses Gulp for their build process and if they can do it for an app of that size, I doubt we would ever have issues.
1. Gulp provides a very clean way of chaining processes together which would easily allow us to add additional tasks or steps as part of gulp tasks.

Thoughts?

timche:
> we could utilize gulp to minify the css

CSS is loaded from the local disk and afaik minifying the CSS doesn't gain us performance. Same can be applied to our JS which is also distributed unminified.

> to use a css preprocessor like SCSS

The question is how many features of SCSS we need. Personally I can live without nested selectors and properties. Mixins are nice, but I think not really essential when defining variables. CSS has support for [global CSS variables](https://developer.mozilla.org/en-US/docs/Web/CSS/:root#Example).

> Without Gulp, npm scripts can get out of hand and hard to manage, which is why Gulp is a nice option.

Don't get me wrong, I agree with npm scripts, on the other hand a build process should only be as complex as the application is and this is where I think Gulp is currently a little too much for our structure/app, where VSCode is a huge and complex app. In comparison, our app is fairly small and just a wrapper to this date. I'm open for Gulp, but only when 1. npm scripts are limiting us 2. we need to chain tools. We should revisit this in a dedicated issue to unblock TS.

Other than that, I like where we are going! :)

markypython:
Sounds good.  I'll work on updating this branch to remove Gulp and address the other items.  I appreciate you taking the time to explain your position as it helps me to get a better understanding of your vision for this app.  Love where we are going and the improvements that have been made so far!

bsramin:
it is not a new project with which to keep the "style" organized well (in scss or less or...) and then process it. I'm overriding on google style, so it's better (imho) to stay on css and have precise pointers.

markypython:
@timche I've updated based on your comments.  Let me know if it needs any additional changes.

package.json
```diff
@@ -25,20 +25,41 @@
...
+    "bracketSpacing": true,
     "useTabs": false,
     "semi": false,
-    "bracketSpacing": true,
     "singleQuote": true
   },
   "xo": {
+    "extends": [
+      "xo-typescript",
+      "prettier/@typescript-eslint"
```
timche: IIRC you don't need to add `prettier/@typescript-eslint` when `prettier: true` is set below. Also see here: https://github.com/sindresorhus/caprine/blob/master/package.json#L59-L73

src/app.ts
```diff
@@ -1,31 +1,33 @@
-const path = require('path')
-const {
+import { join as joinPath } from 'path'
```
timche: I'd like to keep having the default import `path` (or Node APIs in general) instead of using named import, because seeing `joinPath` could lead to thinking it's a custom function and not `path.join`.

src/debug.ts
```diff
@@ -0,0 +1,35 @@
...
+  const state = enabled ? 'enable' : 'disable'
+
+  dialog.showMessageBox(
+    {
+      type: 'info',
+      buttons: ['Restart', 'Cancel'],
+      message: 'Restart required',
+      detail: `To ${state} debug mode, please restart Gmail Desktop`
+    },
+    response => {
```
timche: Type missing?

src/state/window.ts
```diff
@@ -0,0 +1,67 @@
...
+import appConfig from 'electron-settings'
+
+interface State {
+  bounds?: Rectangle
+  isFullScreen?: boolean
+  isMaximized?: boolean
+}
+
+export default class WindowState {
+  private _state: State
```
timche: This is personal preference, but how about omitting the underscore prefix, because the property is already declared as `private`? Same for the rest below.

src/preload.ts
```diff
@@ -0,0 +1,53 @@
+/* eslint-env browser */
```
timche: We should set the `env` in xo instead: https://github.com/sindresorhus/caprine/blob/master/package.json#L64-L67

src/app.ts
```diff
@@ -94,20 +96,20 @@ app.on('ready', () => {
 
   if ((is.linux || is.windows) && !tray) {
     const appName = app.getName()
-    const iconPath = path.join(__dirname, '..', 'static', 'tray-icon.png')
+    const iconPath = joinPath(__dirname, '..', 'resources', 'tray-icon.png')
```
timche:
```suggestion
    const iconPath = joinPath(__dirname, '..', 'static', 'tray-icon.png')
```

timche:
We should also run TS on Travis.

package.json
markypython: Without that, I get the following errors:

`Expected indentation of 3 tabs but found 6 spaces.                 @typescript-eslint/indent`

src/app.ts
markypython: Fixed.  I was thinking along the lines of webpack tree-shaking but that concept isn't in place here.  Mixing up desktop and web apps :)

src/debug.ts
markypython: dialog.showMessageBox provides the type of number without us needing to explicitly state it.

src/state/window.ts
markypython: That was one of the rules in xo TypeScript (`@typescript-eslint/member-naming`).  I can disable if you like.  I would agree that it doesn't look that great.

src/preload.ts
markypython: I didn't want to do this since this is the only file that will have the browser environment.  Unless the others should?

src/app.ts
markypython: Good catch

markypython:
> We should also run TS on Travis.

Not sure I understand what you mean.  Do you mean running something like `tsc --noEmit`?

timche:
Correct. We could have a script `"type-check: "tsc --noEmit"` to do that.

src/preload.ts
timche: I think it's alright to have globally set.

markypython:
Updated based on your feedback.

src/state/window.ts
timche: I just had a quick look for the reasoning to use `_`: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/member-naming.md

> For example, prefixing private properties with a `_` allows them to be easily discerned when being inspected by tools that do not have knowledge of TypeScript (such as most debuggers).

I think it's a reason I can live with.

markypython:
The only other thing I noticed that changed with the new xo rules is that it requires space between member variables (Expected blank line between class members.)  I updated window.ts to fix the rule, but it did look strange to me.

package.json
timche: Can you try to set [`"space": true`](https://github.com/xojs/xo#space) on `xo`?

markypython: Just tried that now and it doesn't change anything

timche:
> it requires space between member variables

Looks ok to me 👍

package.json
timche: Interesting. I'll look into xo and xo-typescript as it seems like a bug.

markypython:
🎉 Thanks so much for your collaboration on this PR!  So glad we made the big update to TypeScript!

PR #42, opened 2019-04-03
More native look&feel

bsramin:
Inspired by https://github.com/fgnass/google-calendar-app


src/app.js
```diff
@@ -55,6 +57,19 @@ function createWindow() {
...
 
   mainWindow.loadURL('https://mail.google.com')
+
+  mainWindow.webContents.on('dom-ready', function() {
+    fs.readFile(__dirname + '/../static/style.css', 'utf-8', function(
+      error,
+      data
+    ) {
+      if (!error) {
+        var formatedData = data.replace(/\s{2,10}/g, ' ').trim()
```
markypython: @timche In #40 I updated to using Gulp for building the app.  This would be a good place for this logic to go as we could use gulp-cssmin to minify the css as part of the prebuild process.

src/app.js
```diff
@@ -55,6 +57,19 @@ function createWindow() {
   WindowState.use('main', mainWindow)
 
   mainWindow.loadURL('https://mail.google.com')
+
+  mainWindow.webContents.on('dom-ready', function() {
+    fs.readFile(__dirname + '/../static/style.css', 'utf-8', function(
```
markypython:
```suggestion
    fs.readFile(path.join(__dirname, '..', 'static', 'style.css'), 'utf-8', (
```

src/app.js
```diff
@@ -55,6 +57,19 @@ function createWindow() {
   WindowState.use('main', mainWindow)
 
   mainWindow.loadURL('https://mail.google.com')
+
+  mainWindow.webContents.on('dom-ready', function() {
```
markypython:
```suggestion
  mainWindow.webContents.on('dom-ready', () => {
```

src/app.js
```diff
@@ -55,6 +57,19 @@ function createWindow() {
...
+    fs.readFile(__dirname + '/../static/style.css', 'utf-8', function(
+      error,
+      data
+    ) {
```
markypython:
```suggestion
    ) => {
```

static/style.css
```diff
@@ -0,0 +1,25 @@
+#gb {
+  padding-left: 60px; /* make room for mac traffic lights */
+  -webkit-app-region: drag; /*make titlebar draggable */
```
markypython:
```suggestion
  -webkit-app-region: drag; /* make titlebar draggable */
```

src/app.js
```diff
@@ -55,6 +57,20 @@ function createWindow() {
...
 
   mainWindow.loadURL('https://mail.google.com')
+
+  mainWindow.webContents.on('dom-ready', () => {
+    fs.readFile(
+      path.join(__dirname, '/../static/style.css'),
+      'utf-8',
+      (error, data) => {
+        if (!error) {
+          let formatedData = data.replace(/\s{2,10}/g, ' ').trim()
```
markypython: @timche In #40 I updated to using Gulp.  I would like to see this whitespace replacement be part of that process using gulp-cssmin.

src/app.js
```diff
@@ -55,6 +57,20 @@ function createWindow() {
   WindowState.use('main', mainWindow)
 
   mainWindow.loadURL('https://mail.google.com')
+
+  mainWindow.webContents.on('dom-ready', () => {
+    fs.readFile(
+      path.join(__dirname, '/../static/style.css'),
```
markypython:
```suggestion
      path.join(__dirname, '..', 'static', 'style.css'),
```

markypython:
Overall, I love where this is going.  I do thing it would be good to wait on this until we get the TypeScript update finished.  That way we can get the css into the prebuild.

src/app.js
markypython:
```suggestion
          const formattedData = data.replace(/\s{2,10}/g, ' ').trim()
```

bsramin:
I modified the css to make the interface more cleaner.

static/style.css
```diff
@@ -0,0 +1,93 @@
...
+}
+
+/* Remove gmail logo */
+.gb_ja:not(.gb_ka) .gb_qc .gb_Rb {
+  display: none !important;
+}
+
+/* Header */
+.gb_lc {
+  height: 32px !important;
```
markypython: I'm assuming we need all these importants to override Gmail styles?  If not, let's remove them.

bsramin: Alas, I tried without important, but they didn't act.

bsramin:
should be tested for windows and linux

src/app.js
```diff
@@ -153,3 +163,16 @@ app.on('activate', () => {
...
+    path.join(__dirname, '..', 'static', 'style.css'),
+    'utf-8',
+    (error, data) => {
+      if (!error) {
+        const formattedData = data.replace(/\s{2,10}/g, ' ').trim()
+        windowElement.webContents.insertCSS(formattedData)
+      }
+    }
+  )
+}
```
timche: Although it's good that you thought of minifying the CSS, imo there is actual no real benefit of doing it and the CSS is loaded from the local disk anyway.

I'd try to keep it as simple as possible like so:
```js
webContents.insertCSS(readFileSync(path.join(__dirname, '..', 'css', 'style.css'), 'utf8'));
```

static/style.css
timche: It's essential to `!important` to either overwrite or protect the value from being overwritten depending on the CSS loading order, so our CSS always has the top priority.

src/app.js
bsramin: Done

bsramin:
let's make an opensoure version of that https://www.boxysuite.com/  ?

timche:
Do we have a way to preview this how it looks like on macOS, Linux and Windows?

markypython:
You could checkout this: https://www.parallels.com.

timche:
@markypython would you do me a favour and could screenshot both?

markypython:
Sure

markypython:
*Mac*

mac

*User icon*

Notice that the focus border is cut off a bit

icon

*Search*

Notice how there is no border or box shadow.

search

*Windows*

Linux looks exactly the same.  Notice how the hamburger menu is offset.

windows

timche:
Thanks a lot @markypython!

> Linux looks exactly the same. Notice how the hamburger menu is offset.

This is what I almost thought. I think we should have separate custom css for macOS/Linux and Window, so the hamburger menu isn't offset. WDYT? @markypython @bsramin

markypython:
Agreed.  We should probably have 4 stylesheets: common, Mac, Windows, Linux.

timche:
I'd split up the CSS file into: 
- `style.css`: Common CSS
- `style.mac.css`: macOS only
- `style.linux.css`: Linux only
- `style.windows.css`: Windows only

timche:
image

I noticed that the avatar overlaps the border below. Can we please make the avatar a little smaller so it doesn't overlap?

css/style.css
```diff
@@ -0,0 +1,93 @@
...
+  background-color: transparent !important;
+  box-shadow: none !important;
+  -webkit-box-shadow: none !important;
+}
+
+/* Main area */
+.G-atb::before {
+  box-shadow: none !important;
+  -webkit-box-shadow: none !important;
+}
```
timche: I didn't really notice any difference with and without this CSS. Where exactly is this changing?

bsramin: this line under the "checkbox"
immagine

more "polish" interface

immagine

bsramin: I want to remove also the line between every single message, but it's really my thing, for now I focus on more important things and then maybe we move on to the dark.

timche: Can we keep this line please? 

I think your interface looks a little different than mine:
image

But anyway I'd like to keep the original interface except customizing it for native looks.

bsramin: ok 👍

src/app.js
```diff
@@ -165,7 +165,20 @@ app.on('before-quit', () => {
 })
 
 function addCustomCss(windowElement) {
+  const cssBaseFile = 'style.css'
```
markypython: Why is this a variable?

src/app.js
```diff
@@ -165,7 +165,20 @@ app.on('before-quit', () => {
...
+
+  if (is.macos) {
```
markypython: This condition is unfortunate.  It would be nice if there was an easier way to handle this, but I do not know of a way to do it currently.

markypython:
The TypeScript changes have been merged, so you will need to merge from master.

src/app.js
bsramin: I don't like it either. "Is" doesn't have a generic verification method for putting it on a switch. Other solutions imho become a useless complication, perhaps more elegant

src/app.js
bsramin: to have all the css not hardcoded in the fs.readFileSync.
Maybe later on it would be more appropriate to better abstract this part.

bsramin:
I've noticed.

bsramin:
Typescript'd

src/app.ts
```diff
@@ -1,4 +1,5 @@
-import path from 'path'
```
markypython: I don't think you need to use the star notation here.  The tsconfig we are using has `allowSyntheticDefaultImports` set to true.

src/app.ts
```diff
@@ -89,6 +96,31 @@ function createMailto(url: string): void {
...
+  const cssBaseFile = 'style.css'
+  let cssOs: string | null = null
+
+  if (is.macos) {
+    cssOs = 'style.mac.css'
+  } else if (is.linux) {
+    cssOs = 'style.linux.css'
+  } else if (is.windows) {
+    cssOs = 'style.windows.css'
+  } else {
```
markypython: Since you are deafaulting cssOs to null, you don't need this else

src/app.ts
```diff
@@ -89,6 +96,31 @@ function createMailto(url: string): void {
...
+    cssOs = null
+  }
+
+  windowElement.webContents.insertCSS(
+    fs.readFileSync(path.join(__dirname, '..', 'css', cssBaseFile), 'utf8')
+  )
+
+  if (cssOs !== null) {
```
markypython: When would this be null?  We are only deploying to Windows, Mac, and Linux, so I wouldn't think this would be necessary.

src/app.ts
```diff
@@ -89,6 +96,31 @@ function createMailto(url: string): void {
   )
 }
 
+function addCustomCss(windowElement: BrowserWindow): void {
+  const cssBaseFile = 'style.css'
```
markypython: Regarding my previous comment about this, I wouldn't extract cssBaseFile since we are only using it in one place.

@timche Thoughts?

css/style.css
```diff
@@ -0,0 +1,96 @@
+#gb {
+  -webkit-app-region: drag; /* make titlebar draggable */
```
markypython: This makes the search box, hamburger menu, and profile picture draggable.  We should revert to the default styles for those elements.

bsramin: what do you mean?

markypython: If you click on the hamburger menu or search box and move your finger or mouse slightly, it will change the position of the window.  It should be setup so that only the blank part of the menu is draggable.

markypython:
@bsramin By the way, thanks so much for you contribution.  It is tough when something like this takes a lot of back and forth, but I really like where we are going with this PR!

src/app.ts
timche: Agree.

src/app.ts
```diff
@@ -89,6 +96,31 @@ function createMailto(url: string): void {
   )
 }
 
+function addCustomCss(windowElement: BrowserWindow): void {
```
timche: I thought of a shorter alternative:

```ts
import * as utils from 'electron-util'

const platform = utils.platform({ macos: 'macos', linux: 'linux', windows: 'windows'})

function addCustomCSS(windowElement: BrowserWindow): void {
  windowElement.webContents.insertCSS(
    fs.readFileSync(path.join(__dirname, '..', 'css', 'style.css'), 'utf8')
  )
  
  const platformCSSFile = `style.${platform}.css`
  if (fs.existsSync(platformCSSFile)) {
    windowElement.webContents.insertCSS(
      fs.readFileSync(path.join(__dirname, '..', 'css', platformCSSFile), 'utf8')
    )
  }
}
```

We don't need to add any other CSS files if they don't have yet imo.

css/style.css
timche: I think that behaviour is actually ok. We'd limit the space where you can drag and I think the times where you accidentally click in the hamburger menu/search bar and drag the window seems quite low to me.

markypython:
@timche The reason I bring up the issue about dragging is I noticed it right away when testing the new window title bar.  In fact, it was quite annoying clicking on the window and causing the window to move around.

bsramin:
Done refactoring of addCustomCSS()

css/style.css
timche: I just tested it and I agree, it's actually very sensitive.

@bsramin can we make the elements in the title bar non-draggable please?

bsramin:
Done!

bsramin:
need an update to README.md for a new screenshot without title bar, can you do that @markypython or @timche ?

timche:
Sure, once this is in merged, I'll update the screenshot.

markypython:
I think the only thing left is to increase the height of the title bar so the avatar focus border doesn't overlap.  @timche Other things you noticed.

When you fix the height, you will probably need to make sure the search bar and hamburger menu are centered vertically in the title bar.

bsramin:
@markypython I already fixed this issue, do you still see it like this?

bsramin:
immagine

markypython:
Here is what I am seeing.

Focus border getting cut off.



Search bar vertical alignment.

bsramin:
ok, I try to fix them

bsramin:
but do not increase the vertical space, I remove the focus, because I can't move the "traffic light" vertically and if I make the bar more higher everything gets misaligned, with a bad effect.

timche:
If we just move the search bar a little higher, so it's vertically centered, it should be fine.

bsramin:
Fixed searchbar and avatar. Check now please.

bsramin:
immagine

markypython:
@timche Do we want to merge this in and release it?

markypython:
v2.4.0 released with this improvement.

timche:
@all-contributors please add @bsramin for code and ideas

bsramin:
thanks @timche

PR #44, opened 2019-04-07
Update changelog

markypython:
Update the changelog for the next release (2.3.0) and add a postversion script to make versioning process easier.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,6 @@
   "main": "dist-js/app.js",
   "scripts": {
     "postinstall": "electron-builder install-app-deps",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -10,6 +10,7 @@
   "main": "dist-js/app.js",
   "scripts": {
     "postinstall": "electron-builder install-app-deps",
+    "postversion": "git push && git push --tags",
```
timche: I'd like to keep these kind of things up to the developer, so in this case it'd better if you set up an alias on your machine, because it can be handy also for other repos.

Issue #47, opened 2019-04-09
Auto updater not working


markypython:
Auto updater was added in #12, but I don't think it is working for Mac since I haven't gotten the new release yet.  I tried restarting the app several times to see if that would trigger it to update, but that doesn't seem to do anything.

@timche Can you explain how auto updater is supposed to work and if there is something I am doing wrong?  Seems like there should be a more intuitive process for updating.

timche:
According to the [documentation](https://www.electron.build/auto-update) and [this example](https://github.com/iffy/electron-updater-example), it should work out of the box when releasing on GitHub and calling `autoUpdater.checkForUpdatesAndNotify()`. Let's see what the log says in the next versions.

timche:
After launching v2.5.0, this is the log:

```
[2019-04-09 13:07:43.017] [info] Checking for update
[2019-04-09 13:07:45.013] [info] Update for version 2.5.0 is not available (latest version: 2.5.0, downgrade is disallowed).
```

It looks like that it should work, let's see when we have released a new version.

markypython:
Behold the culprit:

```
[2019-04-09 20:00:22.126] [info] Checking for update
[2019-04-09 20:00:23.706] [info] Found version 2.5.2 (url: gmail-desktop-2.5.2-mac.zip, gmail-desktop-2.5.2.dmg)
[2019-04-09 20:00:23.707] [info] Downloading update from gmail-desktop-2.5.2-mac.zip, gmail-desktop-2.5.2.dmg
[2019-04-09 20:00:23.730] [info] No cached update info available
[2019-04-09 20:00:23.874] [error] Error: Error: Cannot download "https://github.com/timche/gmail-desktop/releases/download/v2.5.2/gmail-desktop-2.5.2-mac.zip", status 404: Not Found
    at ClientRequest.request.createRequest.response (/Applications/Gmail Desktop.app/Contents/Resources/app.asar/node_modules/builder-util-runtime/src/httpExecutor.ts:247:26)
    at ClientRequest.emit (events.js:182:13)
    at URLRequest.ClientRequest.urlRequest.on (/Applications/Gmail Desktop.app/Contents/Resources/electron.asar/browser/api/net.js:207:12)
    at URLRequest.emit (events.js:182:13)
From previous event:
    at ElectronHttpExecutor.download (/Applications/Gmail Desktop.app/Contents/Resources/app.asar/node_modules/electron-updater/src/electronHttpExecutor.ts:23:73)
[...]

PR #55, opened 2019-04-10
Specify artifact name to fix auto updater issues

markypython:
I think this should fix the update problems we are having in #47.  I haven't tested the output for Windows, but Mac and Linux appear to build the proper file names.  @timche Could you test this for Windows?

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,6 +89,7 @@
     }
   },
   "build": {
+    "artifactName": "${name}-${version}-${os}.${ext}",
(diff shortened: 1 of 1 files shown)


yanlee26:
SQRLUpdaterErrorDomain',
  stack: 'Error: Could not locate update bundle for com.github.Electron within file:///Users/elliotyan/Library/Caches/com.github.Electron.ShipIt/update.uQC09x2/

PR #57, opened 2019-04-10
Remove auto updater logging

markypython:
Now that auto updater is working (#55), the logging can be removed.


timche:
1. I'd like to keep logging in in case for future debugging.
2. `checkForUpdates` is actually what we want to use to silently auto-update the app.

markypython:
Okay, never mind this PR then!  Thanks for the explanation.

Issue #62, opened 2019-04-12
Close window on click "Archive" when email opened in new window


timche:
When opening an email via an notification, the email will be opened in a new window. It would be nice when clicking on "Archive" the window will be closed automatically. This is a behaviour I know from the Gmail app on iOS.

This could be made configurable.

markypython:
Same behavior when you click delete as well

timche:
There is also a "Report spam" button, could be the same behaviour.

PR #58, opened 2019-04-11
Fix new window header

markypython:
Fixes #50

PR #69, opened 2019-04-12
Open links from new windows in an external browser

### LINKED ISSUE #60: External links open in new window from new window
- Gmail Desktop Version: latest
- OS Version: macOS Mojave

Steps to Reproduce:

1. Ctrl/cmd click on an email to open it in a new window
2. Click on an external link in the email (i.e. GitHub PR)
3. See how the link was opened in a new window, should be opened using `shell.openExternal` to open in a browser

markypython:
Fixes #60

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -183,9 +183,18 @@ app.on('ready', () => {
         x: null,
         y: null
       })
+
       event.newGuest.webContents.on('dom-ready', () => {
         addCustomCSS(event.newGuest)
       })
+
+      event.newGuest.webContents.on(
+        'new-window',
+        (event: Event, url: string) => {
+          event.preventDefault()
+          shell.openExternal(url)
+        }
+      )
     } else {
       shell.openExternal(url)
     }
--- a/src/preload.ts
+++ b/src/preload.ts
@@ -1,6 +1,8 @@
 import { ipcRenderer as ipc } from 'electron'
 
-const INTERVAL = 1000
+const UNREAD_INTERVAL = 1000
+const ATTACH_BUTTON_INTERVAL = 200
+
 let count: number
 
 function getUnreadCount(): number {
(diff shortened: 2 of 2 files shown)


markypython:
Also fixes #62

markypython:
Requires the electron downgrade in PR #58

timche:
Please merge this PR rebased to keep the commits and don't squash them as this PR does multiple things, not only the PR title.

markypython:
The `load` event fires before the buttons are created on the page.  In normal circumstances, it will find the button the first time, but it is possible in scenarios where the network is slow for it to take a little while.

I can update to element-ready if you would like, but I would recommend we wait on that until we have future needs for this functionality.  Then we can look into revamping it.

Thoughts?

timche:
> I would recommend we wait on that until we have future needs for this functionality

With using `element-ready` I think we can improve the logic and make our code more readable and stable. I can also re-write it if you want.

markypython:
So, I looked into the source of element-ready and it is using `requestAnimationFrame` to check if the element is present on the page.  This seems like overkill for what we need since that would be checking 60 times per second if the element is on the page.  I still think we should leave it as is, but if you disagree, let me know.

timche:
I'm open to discuss the pros and cons. According to this [issue](https://github.com/sindresorhus/element-ready/issues/4), performance-wise it shouldn't be that CPU-intensive. Also using `setTimeout` can still leave a very small time-window open where the listener is not attached when clicked before, although the change might be quite low, but it can happen.

timche:
By the way, I tried to checkout this PR, but the buttons aren't working for me. Anything I'm missing? I used the "In new window" button to open an email in a new window.

image

timche:
Would you mind creating a separate PR for `Open links from new windows in an external browser` so we can get that in while we are discussing button listeners?

markypython:
I'll open two new PRs for the separate branches.

PR #70, opened 2019-04-13
Minimal mode

### LINKED ISSUE #59: "Minimal mode" idea
I was thinking about possibly adding a "minimal mode" option that would hide certain elements on the page such as the right side panel, the support question mark, the footer information, keyboard button, etc.

@timche Thoughts on adding this as an option?

markypython:
Adds a new minimal mode to clean up some of the messy Gmail interface.  Fixes #59

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,9 +1,12 @@
 import { BrowserWindow } from 'electron'
 
-export function getWindow(): BrowserWindow {
+export function getMainWindow(): BrowserWindow {
   return BrowserWindow.getAllWindows()[0]
 }
 
-export function sendAction(action: string, ...args: unknown[]): void {
-  getWindow().webContents.send(action, ...args)
+export function sendChannelToMainWindow(
+  channel: string,
+  ...args: unknown[]
+): void {
+  getMainWindow().webContents.send(channel, ...args)
 }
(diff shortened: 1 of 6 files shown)


src/utils.ts
```diff
@@ -0,0 +1,9 @@
+import { BrowserWindow } from 'electron'
+
+export function getWindow(): BrowserWindow {
+  return BrowserWindow.getAllWindows()[0]
```
timche: The first window is the main window where `gmail.com` is loaded right? Should we name the function `getMainWindow()` instead?

src/utils.ts
```diff
@@ -0,0 +1,9 @@
...
+}
+
+export function sendAction(action: string, ...args: unknown[]): void {
```
timche: How about naming it more specific like `sendChannelToMainWindow()` to use the same term as in the [doc](https://electronjs.org/docs/api/web-contents#contentssendchannel-arg1-arg2-)?

timche:
Please merge this PR rebased to keep the commits and don't squash them as this PR does multiple things, not only adding the minimal mode.

timche:
I'm excited for the new update 😃

src/utils.ts
markypython: Yes, the first window is the one loaded when the app starts.  I will rename the function

src/utils.ts
markypython: Updated!

PR #81, opened 2019-04-14
use `config.dev.json` for development

timche:


markypython:
When I setup debug mode, I specifically didn't want to have it done this way because that makes it very difficult to test if the feature is working.  I did however set it up to default to enabled when in the development environment (second argument to `appConfig.get`).

timche:
The problem is that this setting is shared between the development and production app, so whether I enable it in one, it's also enabled in the other. I think there is not too much to worry if this feature is working or not, because we have TypeScript and it's not much code. WDYT?

markypython:
I wonder if when we switch to electron-store if we could use a store name that includes the environment to prevent the shared config between environments, but still maintain the ability to turn off debug mode for development.

timche:
Yes, it is possible to define the name of the storage file: https://github.com/sindresorhus/electron-store#name

timche:
Currently blocked by #105

timche:
@markypython I've changed this PR to "use `config.dev.json` for development". Can you review again please?

PR #84, opened 2019-04-14
Close new window when email is archived, marked as spam, or deleted

### LINKED ISSUE #62: Close window on click "Archive" when email opened in new window

markypython:
Fixes #62

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,8 @@
     "electron-log": "^3.0.5",
     "electron-settings": "^3.2.0",
     "electron-updater": "^4.0.6",
-    "electron-util": "^0.11.0"
+    "electron-util": "^0.11.0",
+    "element-ready": "^3.1.0"
   },
   "devDependencies": {
     "@sindresorhus/tsconfig": "^0.3.0",
--- a/src/app.ts
+++ b/src/app.ts
@@ -183,6 +183,7 @@ app.on('ready', () => {
         x: null,
         y: null
       })
+
       event.newGuest.webContents.on('dom-ready', () => {
         addCustomCSS(event.newGuest)
       })
--- a/src/preload.ts
+++ b/src/preload.ts
@@ -48,7 +48,6 @@ function attachButtonListeners(): void {
     .then(() => {
       const selectors = [
         'lR', // Archive
-        'nN', // Spam
         'nX' // Delete
       ]
 
--- a/src/preload.ts
+++ b/src/preload.ts
@@ -1,8 +1,8 @@
 import { ipcRenderer as ipc } from 'electron'
+import elementReady from 'element-ready'
+import log from 'electron-log'
 
-const UNREAD_INTERVAL = 1000
-const ATTACH_BUTTON_INTERVAL = 200
-
+const INTERVAL = 1000
 let count: number
 
 function getUnreadCount(): number {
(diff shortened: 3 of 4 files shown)


timche:
Follow up from #69:

---

**1. Testing Question:**

I tried to checkout this PR, but the buttons aren't working for me. Anything I'm missing? I used the "In new window" button to open an email in a new window.

image

---

**2. Discussion about using `element-ready` package instead of `setTimeout`:**

@markypython:
> So, I looked into the source of element-ready and it is using `requestAnimationFrame` to check if the element is present on the page.  This seems like overkill for what we need since that would be checking 60 times per second if the element is on the page.  I still think we should leave it as is, but if you disagree, let me know.

@timche
> I'm open to discuss the pros and cons. According to this [issue](https://github.com/sindresorhus/element-ready/issues/4), performance-wise it shouldn't be that CPU-intensive. Also using `setTimeout` can still leave a very small time-window open where the listener is not attached when clicked before, although the change might be quite low, but it can happen.

markypython:
I refactored to use element-ready and added a 10 second safety net in case element-ready starts in a window that doesn't have the buttons we are looking for.  Additionally, I remove the spam button because it requires a prompt after clicking the button before the message is marked as spam.

markypython:
> I tried to checkout this PR, but the buttons aren't working for me. Anything I'm missing? I used the "In new window" button to open an email in a new window.

I didn't notice this behavior.  Could you open a separate issue for this so we can discuss it futher?

timche:
> I didn't notice this behavior. Could you open a separate issue for this so we can discuss it further?

Can you share a test plan just in case?

markypython:
I tested on all platforms and it works fine.  Can you try reinstalling yarn packages?  This requires the electron downgrade from one of the other PRs so that might be the issue you are experiencing.

timche:
I reinstalled all dependencies. Still doesn't work for me.

Can you share with me your test steps? Not sure if I'm just doing something wrong.

markypython:
My testing steps were fairly simple.

1. Checkout the `close-new-window` branch
1. Run `yarn && yarn start`
1. Ctrl/cmd click on an email to open it in a new window
1. Click the archive or delete button and observe the new window being closed
1. Repeat 3-4 but using the open in new window button you showed in your screenshot

markypython:
The two things you removed, `if (button) { ...` and the `setTimeout` were both added for strategic reasons which I will explain below.

*if button*

`document.querySelector` can return null and therefore having a check to verify the button is there is wise.  If Gmail selectors changed in the future, this would cause errors.  The likelihood that the selectors will change is probably almost nothing, so I would probably be okay with changing it to `button!.addEventListener`

*Timeout*

This one is more important in my opinion.  I Gmail url schemes change were our conditional no longer works exactly how we want it to, it could be possible for element ready to run in a window endlessly.  I'm of the opinion that the small bit of extra code for putting a safeguard like this in place is better than the possibility of creating a runaway check loop.  Yes, it might not be bad for performance, but running document.querySelector 60 times per second for 10 minutes might not be great for performance if such a case were to occur.

Thoughts?

timche:
Thanks for sharing your thoughts!

---

**if button**

That would be solved waiting for each selector to be ready: `elementReady(`body.xE .G-atb .${selector}`)`

**Timeout**

Convinced.

---

I've added a timeout to my previous example:

```ts
function attachButtonListeners(): void {
  // For windows that won't include the selectors we are expecting,
  //   don't wait for them appear as they never will
  if (!window.location.search.includes('&search=inbox')) {
    return
  }

  const selectors = [
    'lR', // Archive
    'nX' // Delete
  ]

  selectors.forEach(async selector => {
    try {
      const buttonReady = elementReady(`body.xE .G-atb .${selector}`)

      const readyTimeout = setTimeout(() => {
        buttonReady.cancel(`Detect button "${selector}" timed out`)
      }, 10000)

      const button = await buttonReady
      clearTimeout(readyTimeout)

      button.addEventListener('click', () => window.close())
    } catch (err) {
      log.error(err)
    }
  })
}
```

My goal is to use  `async/await` instead of promise chaining to make the code more readable and modern. Additionally instead to wait for one button to access all others, wait for each button to make fully sure that it actually exists, so we don't have to double-check with `document.querySelector()` and `if (button)`. If one button timeouts, `cancel()` the promise and log the cancel error in the `catch` block.

markypython:
Love the snippet you shared.  I'll get it updated today or tomorrow.

markypython:
I updated with the changes you requested.

PR #98, opened 2019-04-19
disable windows code signature check on update

timche:
Same issue on Windows as explained here: https://github.com/sindresorhus/caprine/issues/752#issuecomment-458648468

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,6 +105,9 @@
         "deb"
       ],
       "category": "Network;Office"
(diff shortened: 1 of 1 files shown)


markypython:
Could you explain the reason for this?  I'm sure it is fine, but I just would like to understand why this is being added.

timche:
I forgot to add a description. See top description again.

PR #99, opened 2019-04-19
change app id

timche:
I've registered https://timche.dev for my personal website and want to use this as app id.

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
     "files": [
       "**/*"
     ],
-    "appId": "io.cheung.gmail-desktop",
+    "appId": "dev.timche.gmail-desktop",
     "mac": {
       "category": "public.app-category.productivity",
       "darkModeSupport": false
--- a/src/app.ts
+++ b/src/app.ts
@@ -33,7 +33,7 @@ if (!is.development) {
   autoUpdater.checkForUpdates()
 }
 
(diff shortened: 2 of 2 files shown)


markypython:
Will this affect auto updating to the new app id?

timche:
Good question, I will test this.

timche:
I couldn't really test auto-update, but I generated a `.dmg` on macOS and installed it over the old version. Seems to work like before.

macOS:

> CFBundleIdentifier (String - iOS, macOS) uniquely identifies the bundle. Each distinct app or bundle on the system must have a unique bundle ID. The system uses this string to identify your app in many ways. For example, the preferences system uses this string to identify the app for which a given preference applies; Launch Services uses the bundle identifier to locate an app capable of opening a particular file, using the first app it finds with the given identifier; in iOS, the bundle identifier is used in validating the app’s signature.

Windows:

> Application User Model IDs (AppUserModelIDs) are used extensively by the taskbar in Windows 7 and later systems to associate processes, files, and windows with a particular application. In some cases, it is sufficient to rely on the internal AppUserModelID assigned to a process by the system. However, an application that owns multiple processes or an application that is running in a host process might need to explicitly identify itself so that it can group its otherwise disparate windows under a single taskbar button and control the contents of that application's Jump List.

I don't think that it will affect the auto-update in any way and it's just used to identify the process(es) itself on system level.

PR #104, opened 2019-04-25
rename app to `Mailtron` and change icon

### LINKED ISSUE #39: Rebrand project
As we are using Gmail and it's logo, it can potentially happen that we are getting a legal notice as it could confuse people that it looks like an official app and we are asked to change the name and logo. In order to avoid these issues in advance, we could rename our app, but I'm no legal expert if we actually need to.

timche:
- I opened this PR as a draft to give us some more time to finalize and decide on this renaming.
- Feel free to checkout this PR and build it on your machine.
- Unfortunately renaming the whole project will also change the user data path, so all settings will be gone.
  - We are still at the very beginning and there is not much to configure, so I think it's fine apart from forcing everyone to re-login.
  - We can avoid this by changing `productName` in `build.productName` instead and leave everything as is, but this feels kinda messy.
- I think auto-updating won't work anymore after this, so everyone needs to manually update/re-install this app.

This is how the icon looks like in the dock:

Screenshot 2019-04-25 at 14 09 38

Closes #39


package.json
```diff
@@ -90,11 +90,11 @@
     }
   },
   "build": {
-    "artifactName": "${name}-${version}-${os}.${ext}",
     "files": [
-      "**/*"
+      "**/*",
+      "!media${/*}"
```
markypython: What is this change for?

timche: So the `media` folder won't be included in the build.

timche:
> Overall, looks good. What is the icon.fig file for? I'm not familiar with that extension.

It's a [Figma](https://www.figma.com) project file, but any app which can open and edit `.svg` can be used too instead of Figma.

markypython:
> It's a Figma project file

Okay, so it's like a .ai or .psd file.  The source used for generating the images.  Right?

timche:
Correct.

timche:
I don't have Illustrator or any other vector program, so as I said, any program which can edit `.svg` can generate the `.png` too.

markypython:
To be clear, I wasn't trying to say we need to have a .ai or .psd file, I just wanted to make sure I understood the purpose of the .fig file.  All good! 👍

timche:
I'm yet not 100% sure how we can do this change and migrate everyone over, as it is practically a new app and manual steps are required. Didn't find any docs/guides how an app can be fully renamed with `electron-builder`.

PR #105, opened 2019-04-29
replace `electron-settings` with `electron-store`

### LINKED ISSUE #80: Use electron-store instead of electron-settings
Differences:
- TypeScript support
- [Set defaults](https://github.com/sindresorhus/electron-store#defaults)
- [Define schema](https://github.com/sindresorhus/electron-store#schema)
- Can be easily used in main and renderer process

### LINKED ISSUE #86: `this._state.bounds` can be undefined and crashes the app
image

- Gmail Desktop Version: 2.6.1
- OS Version: Windows 10

Steps to Reproduce:

Not clear how to reproduce exactly, but the error can be reproduced programmatically:

```diff
diff --git a/src/state/window.ts b/src/state/window.ts
index a80809d..e20ddf9 100644
--- a/src/state/window.ts
+++ b/src/state/window.ts
@@ -49,6 +49,7 @@ export default class WindowState {
     } else if (this._state.isMaximized) {
       this._window.maximize()
     } else {
+      delete this._state.bounds
       this._window.setBounds(this._state.bounds!)
     }
   }
```

This happened to me only on Windows (once), but I think this can happen on all OS. It's not safe to use `this._window.setBounds(this._state.bounds!)` when `this._state.bounds` is not defined.

**Possible solution:**
Define window state default

```ts
{
  isMaximized: true,
  isFullscreen: false,
  bounds: {
    width: 800,
    height: 600,
    // Center the window
    y: undefined as unknown as number,
    x: undefined as unknown as number
  }
}
```

timche:
Closes #80 and fixes #86.


src/app.ts
```diff
@@ -61,18 +61,26 @@ app.on('second-instance', () => {
 })
 
 function createWindow(): void {
+  const lastWindowState = config.get('lastWindowState') as LastWindowState
```
timche: For some reason I need to assert the type here because
1. `lastWindowState` will be typed as `boolean | LastWindowState`
2. referencing `lastWindowState.bounds` will throw an error that `bounds` doesn't exist on `boolean`

It's a weird error and I don't know if it's by design, but we should check with the [`electron-store`](electron-store) guys.

This is done everywhere where `config.get()` is called.

timche: Okay, I think this is actually by design and you must assert the type.

timche:
Anything blocking here @markypython? I'd like to bring this in to add other features :)

src/app.ts
timche: Rolling back to `electron-store@3.1.0` without type definitions and using `@types/electron-store` has fixed this behaviour and `config.get()` is now returning the correct type based on the key. I'll check the type definition in `electron-store` and see if I can contribute back.

src/app.ts
```diff
@@ -15,10 +15,10 @@ import log from 'electron-log'
 import electronDl from 'electron-dl'
 import electronContextMenu from 'electron-context-menu'
 
+import config from './config'
 import { init as initDebug } from './debug'
 import menu from './menu'
 import { init as initMinimalMode } from './minimal-mode'
-import WindowState from './state/window'
```
markypython: I'm a bit concerned with moving all of the state logic into app.ts as this file will continue to grow huge and be somewhat unruly.  That is why I was trying to put all the window state logic in the separate file.

src/config.ts
```diff
@@ -0,0 +1,58 @@
...
+  if (oldConfig.has('debug-mode')) {
+    const debugMode = (oldConfig.get('debug-mode') as unknown) as boolean
+    config.set('debugMode', debugMode)
+    oldConfig.delete('debug-mode')
+  }
+
+  if (oldConfig.has('minimal-mode')) {
+    const minimalMode = (oldConfig.get('minimal-mode') as unknown) as boolean
+    config.set('minimalMode', minimalMode)
+    oldConfig.delete('minimal-mode')
```
markypython: Since you are unlinking the old config, do you need to delete the individual settings as well?

markypython:
Overall looks good, just a few questions.

src/app.ts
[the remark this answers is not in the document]
timche: Although you marked this as resolved, my reasoning not doing it this way was to specifically reference `lastWindowState` as value, so it's easier to see where the values are coming from.

src/app.ts
timche: I think we are fine for now, because it's actually much less code/logic than before. If we see this becoming an issue (`app.ts` being too complex), especially when supporting multiple windows in the future, I'd reconsider splitting this file for sure.

src/config.ts
timche: I added `fs.unlinkSync(oldConfigFile)` today, so yeah, this is not required 😅

markypython:
Looks good.  As a note for next time, force pushing prevents me from seeing what you changed after pr feedback, so if you could add a commit and then do a normal push to the branch that would make it easier.

👍

PR #106, opened 2019-04-29
fix account switching

timche:
Currently when adding a new or switching to another account their links will be opened in a new window instead in the main window.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -10,13 +10,14 @@ import {
   MenuItemConstructorOptions
 } from 'electron'
 import { autoUpdater } from 'electron-updater'
-import utils, { is } from 'electron-util'
+import { is } from 'electron-util'
 import log from 'electron-log'
 
 import { init as initDebug } from './debug'
 import menu from './menu'
 import { init as initMinimalMode } from './minimal-mode'
 import WindowState from './state/window'
+import { platform, getUrlAccountId } from './helpers'
 
 // Initialize the debug mode handler when starting the app
 initDebug()
@@ -39,12 +40,6 @@ let mainWindow: BrowserWindow
 let replyToWindow: BrowserWindow
 let isQuitting = false
 let tray: Tray
-
-const platform = utils.platform({
-  macos: 'macos',
-  linux: 'linux',
-  windows: 'windows'
-})
 
 if (!app.requestSingleInstanceLock()) {
   app.quit()
@@ -176,7 +171,19 @@ app.on('ready', () => {
   webContents.on('new-window', (event: any, url, _1, _2, options) => {
     event.preventDefault()
 
-    if (/^(https:\/\/(mail|accounts)\.google\.com).*/.test(url)) {
+    // `Add account` opens `accounts.google.com`
+    if (/^https:\/\/accounts\.google\.com/.test(url)) {
+      mainWindow.loadURL(url)
+    } else if (/^https:\/\/mail\.google\.com/.test(url)) {
+      // Check if the user switches accounts which is determined
+      // by the URL: `mail.google.com/mail/u/<local_account_id>/...`
+      const currentAccountId = getUrlAccountId(mainWindow.webContents.getURL())
+      const targetAccountId = getUrlAccountId(url)
+
+      if (targetAccountId !== currentAccountId) {
+        return mainWindow.loadURL(url)
+      }
+
       // Center the new window on the screen
       event.newGuest = new BrowserWindow({
         ...options,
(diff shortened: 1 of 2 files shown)


src/app.ts
```diff
@@ -176,7 +176,24 @@ app.on('ready', () => {
   webContents.on('new-window', (event: any, url, _1, _2, options) => {
     event.preventDefault()
 
-    if (/^(https:\/\/(mail|accounts)\.google\.com).*/.test(url)) {
+    if (/^https:\/\/accounts\.google\.com/.test(url)) {
+      mainWindow.loadURL(url)
+    } else if (/^https:\/\/mail\.google\.com/.test(url)) {
```
markypython: Can you explain this logic to me?  It seems like when you switch accounts you shouldn't have any urls that point to the old account anymore?

timche: Do mean L181 or L179 + L181?

markypython: L181+

timche: Don't know if I fully understand your question, but I'll try:

The first account you are signing in to is the default account. Every other will be an additional account. The default account is identified as `0` in `https://mail.google.com/mail/u/0/`. When you are switching accounts, the second account for example is `1` in ``https://mail.google.com/mail/u/<id>/`.

markypython: I guess my question is more of when do links fall into this else statement? Which type of links are they? Notifications? mailto type links?

timche: Hm, I can't name all, but at least single email view, compose, pretty much everything(?). I only know of `accounts.google.com` being opened when you trying to add a new account.

timche:
Do you have any concerns regarding these changes?

markypython:
@timche Yesterday was busy for me.  I would like to pull this locally and experiment a little to understand it better.  Do you mind waiting till I take a look at that tonight?

timche:
Sure thing!

src/app.ts
markypython: So, I played around with this and I have a more specific question.  When is this conditional true: `currentAccountId !== accountId`.  It seems like it never will be since all the links in the app will point to the current account.  So my question is are we adding this logic and it will never end up doing anything.

timche: Actually it does, maybe the code isn't clear enough. You need to be logged in with two accounts already.

Current behaviour:

Kapture 2019-05-01 at 0 39 16

Expected behaviour (this PR):

Kapture 2019-05-01 at 0 40 52

markypython: That part I understand, but that is handled by this block

```ts
if (/^https:\/\/accounts\.google\.com/.test(url)) {
      mainWindow.loadURL(url)
}
```

I'm asking if this block is needed

```ts
} else if (/^https:\/\/mail\.google\.com/.test(url)) {
      const accountIdRegExp = /mail\/u\/(\d+)\//

      const currentAccountRes = accountIdRegExp.exec(
        mainWindow.webContents.getURL()
      )
      const currentAccountId = currentAccountRes && currentAccountRes[1]

      const accountRes = accountIdRegExp.exec(url)
      const accountId = accountRes && accountRes[1]

      if (currentAccountId !== accountId) {
        console.log(currentAccountId)
        console.log(accountId)

        mainWindow.loadURL(url)
        return
      }
```

timche: No, that's what I actually tried to explain 😄 

I've `console.log(url)` for you to show you what the URL for which action:

Kapture 2019-05-01 at 0 50 44

See how the url is `https://mail.google.com/mail/u/1` for the second account. It will be `/u/0` for the default account when you switch back. `Add account` opens `https://accounts.google.com`.

Hope that clarifies it.

markypython: Yes, that helps.  Sorry for the back and forth, but I thought it was the first if statement that would fire when switching accounts.

src/app.ts
```diff
@@ -176,7 +176,24 @@ app.on('ready', () => {
...
+      const accountIdRegExp = /mail\/u\/(\d+)\//
+
+      const currentAccountRes = accountIdRegExp.exec(
```
markypython: Maybe pull this logic into a function for getting the account id from a url?

src/app.ts
```diff
@@ -176,7 +176,24 @@ app.on('ready', () => {
...
+        mainWindow.webContents.getURL()
+      )
+      const currentAccountId = currentAccountRes && currentAccountRes[1]
+
+      const accountRes = accountIdRegExp.exec(url)
+      const accountId = accountRes && accountRes[1]
+
+      if (currentAccountId !== accountId) {
+        mainWindow.loadURL(url)
```
markypython:
```suggestion
        return mainWindow.loadURL(url)
```

src/app.ts
timche: No worries, I think I should add a comment to explain the logic.

timche:
🎉

Issue #74, opened 2019-04-13
Add context menu


timche:
Right-clicking on elements and input fields doesn't do anything yet. It should instead show a context menu to do various things like `Copy link`, etc. (TBD) using [electron-context-menu](https://github.com/sindresorhus/electron-context-menu).

cipriancraciun:
At the moment copy-link is essential, especially for those that use multiple browsers / browser profiles.

For example say I receive an email with an alert from a cloud provider, with a link that contains more details, or an action to be taken.  If I use a different browser profile (or even browser instance) to manage my AWS account, if I click that link it opens in my default browser, which due to how the web-application is implemented redirects me to a login page, and my original link is (at best) obscured or (at worst, which is usually) lost.

---

Fortunately in my case I have a custom `xdg-open` script that allows me to intercept the external links, however for the normal user getting to the original link could be an ordeal.

(In fact having this custom `xdg-open` script is another reason what prompted me to switch from the browser to the Electron based application, because I can open links in different browsers / profiles based on their domain.)

markypython:
Added in [v2.8.0](https://github.com/timche/gmail-desktop/releases/tag/v2.8.0)

PR #107, opened 2019-04-29
enhance download experience + add context menu

### LINKED ISSUE #73: Improve download attachments experience
Using [electron-dl](https://github.com/sindresorhus/electron-dl) to give the app even more native feeling.

timche:
Closes #73 and #74

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -12,6 +12,7 @@ import {
 import { autoUpdater } from 'electron-updater'
 import utils, { is } from 'electron-util'
 import log from 'electron-log'
+import electronDl from 'electron-dl'
 
 import { init as initDebug } from './debug'
 import menu from './menu'
@@ -20,6 +21,8 @@ import WindowState from './state/window'
 
 // Initialize the debug mode handler when starting the app
 initDebug()
+
+electronDl({ showBadge: false })
 
 if (!is.development) {
   log.transports.file.level = 'info'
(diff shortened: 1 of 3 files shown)


src/app.ts
```diff
@@ -133,12 +140,14 @@ function addCustomCSS(windowElement: BrowserWindow): void {
   }
 }
 
-app.on('ready', () => {
-  createWindow()
+;(async () => {
```
markypython: Why this change?

src/app.ts
```diff
@@ -12,6 +12,8 @@ import {
 import { autoUpdater } from 'electron-updater'
 import utils, { is } from 'electron-util'
 import log from 'electron-log'
+import electronDl from 'electron-dl'
+import electronContextMenu from 'electron-context-menu'
```
markypython: Does this conflict with the context menu in gmail when right clicking on an email?

src/app.ts
timche: Actually not really needed. I just saw in the `electron-dl` example that `app.whenReady()` can be used as a convenient alternative. I can revert this.

src/app.ts
timche: Can confirm that it doesn't conflict.

src/app.ts
markypython: Yeah, I think the `app.on('ready', () => ...)` is cleaner and fits better with the rest of the app.

timche: Reverted, thanks for bringing this up.

markypython:
I just realized that this plugin shows a badge count for downloads.  We should probably disable that since the badge is already used for unread email count.

https://www.npmjs.com/package/electron-dl#showbadge

timche:
Good catch, fixed.

PR #112, opened 2019-05-01
fix search bar styles

timche:
Gmail shipped an update which changed the search bar and its classes.

Before:
Screenshot 2019-05-01 at 12 41 05

After:
Screenshot 2019-05-01 at 13 02 48


markypython:
I also just noticed that the support question mark class changed.  I committed to this branch to update the class.

PR #117, opened 2019-05-02
Fix fullscreen

markypython:
The fullscreen option of browser window doesn't actually set fullscreen but enables/disables it from being used.


timche:
That's weird, because the description says `Whether the window should show in fullscreen.` 🤔 Anyway, thanks!

Issue #121, opened 2019-05-23
Rethink Custom Styles


timche:
As of now, it's very difficult to keep up with CSS class changes which results in breaking our custom styles constantly. We should come up with a compromise where we apply "safe" custom styles that are breaking rarely.

Opinion @markypython ?

markypython:
@timche I've thought about how to do the custom styles better and I think it would require using fewer class names and more selector hierarchy.  Styles like this `#gb > div > div:last-of-type > span`.

PR #135, opened 2019-08-27
Update custom styles

markypython:
Hey Tim, long time no speak!  I hope all is well.

I switched back to using gmail desktop a month or two ago and I wanted to contribute some more to the project.  The first thing that was bugging me was the custom styles not working.  As I mentioned in https://github.com/timche/gmail-desktop/issues/121, I reworked the custom styles to be based on element position rather than class names.  We can see how it works but I am hoping it will be much more stable than the class name approach.  Additionally, I commented on every line of the stylesheet as there were styles in the old stylesheet that I had no idea why they were there because the class names have since changed.  By keeping every style commented, we can always fix the selectors because we know the intent of the style.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -1,161 +1,105 @@
-/* Make titlebar draggable */
-#gb,
-/* Message toolbar in new window */
-body.xE .G-atb,
-/* Compose in new window */
-.Io,
-/* PDF viewer */
-.aLF-aPX-aPU-J6 {
+/**
+ * DRAGGABLE HEADERS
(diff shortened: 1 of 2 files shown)


timche:
Welcome back @markypython! I've been very busy recently and couldn't find any time to improve Gmail Desktop unfortunately. I really appreciate your contributions!

If you could fix the CI errors, I'm more than happy to merge this PR :)

timche:
Awesome, thank you!

PR #136, opened 2019-08-28
More custom style fixes

markypython:
Fixes a number of issues with the recent custom style changes.

- Adds selectors for dragging headers for new windows and pdf viewer
- Adds left padding for macos for new window and pdf viewer headers
- Fix height for G Suite account links (it is larger than the normal account avatar)


timche:
That's awesome, thanks so much!

PR #138, opened 2019-08-29
Custom style settings

### LINKED ISSUE #108: Make minimal mode enabled by default
@markypython I thought of enabling the minimal mode by default, but also giving the user the option to opt-out of the things that minimal mode is hiding via `Settings > Appearance >`
- `Hide Footer`
- `Hide Right Sidebar`
- `Hide Support` (The questionmark icon on the top right)

markypython:
This adds three additional settings for custom styles to allow finer control of hiding elements.  More details are explains in #108.

I have no migration function in this PR to migration from the old settings to these new settings as I wanted to get this out there to get your feedback so I was sure that the rest of the code isn't changing.

I also added a `ConfigKey` enum so that we don't have a bunch of magic strings floating around for the config keys.

Fixes #108


timche:
Awesome work as always, really appreciate it someone is taking such good care 👍 💯

markypython:
@timche As mentioned in my comment, I hadn't added migration to this PR.  Please don't release again until I can add that migration from the old minimal mode to the new one.

timche:
Oh, alright. Misunderstood that part, sorry. Waiting for another PR then.

markypython:
@timche I was thinking a bit more about the migration strategy for minimal mode and I think the best option is to just release without a migration.  The reason for that is the description of #108 is turn on minimal mode by default.  Since we don't know which users explicitly tried and then turned off minimal mode, we don't know if we should leave it turned off or turn it on as the new default.

Does that make sense?

timche:
Sure, fine for me, thanks!

PR #139, opened 2019-08-29
Fix minor style issues

markypython:
Fix a couple minor style issues.

1. The search box text was moving 1px after Gmail fully loaded
1. The user avatar for G Suite accounts was not being set to no-drag


timche:
Since #138 has been merged, you might want to fix the conflicts. Thanks for your work.

markypython:
@timche Updated!

PR #140, opened 2019-08-30
Update dependencies

markypython:
Removes electron-settings now that it isn't used and updates all dependencies to their latest version.


src/app.ts
```diff
@@ -13,14 +13,15 @@ import { autoUpdater } from 'electron-updater'
...
-import electronContextMenu from 'electron-context-menu'
 
-import config, { ConfigKey } from './config'
+import config, { ConfigKey, LastWindowState } from './config'
 import { init as initDebug } from './debug'
 import menu from './menu'
 import { init as initCustomStyles } from './custom-styles'
 import { platform, getUrlAccountId } from './helpers'
 
+import electronContextMenu = require('electron-context-menu')
```
timche: `import from`

src/app.ts
```diff
@@ -1,5 +1,5 @@
-import path from 'path'
-import fs from 'fs'
+import { readFileSync, existsSync } from 'fs'
```
timche: We had this before and I'd like to keep this as `import fs from 'fs'` as it's more common to use the module name for nodejs apis.

src/app.ts
```diff
@@ -1,5 +1,5 @@
...
+import { join } from 'path'
```
timche: Same as above.

src/config.ts
```diff
@@ -1,6 +1,7 @@
-import Store from 'electron-store'
 import { is } from 'electron-util'
 
+import Store = require('electron-store')
```
timche: `import from`

src/debug.ts
```diff
@@ -1,12 +1,12 @@
-import electronDebug from 'electron-debug'
 import config, { ConfigKey } from './config'
 
+import electronDebug = require('electron-debug')
```
timche: `import from`

src/preload.ts
```diff
@@ -1,9 +1,10 @@
 import { ipcRenderer as ipc } from 'electron'
-import elementReady from 'element-ready'
 import log from 'electron-log'
 
 import { ConfigKey } from './config'
 
+import elementReady = require('element-ready')
```
timche: `import from`

src/preload.ts
```diff
@@ -49,21 +50,19 @@ function attachButtonListeners(): void {
...
-
-      const readyTimeout = setTimeout(() => {
-        buttonReady.cancel(`Detect button "${selector}" timed out`)
-      }, 10000)
+  selectors.forEach(selector => {
+    const buttonReady = elementReady(`body.xE .G-atb .${selector}`)
+    const readyTimeout = setTimeout(() => buttonReady.stop(), 10000)
 
-      const button = await buttonReady
+    buttonReady.then(button => {
```
timche: Why using chaining instead of async/await?

src/app.ts
markypython: This doesn't work as they changed the export to commonjs.

https://github.com/sindresorhus/electron-context-menu/releases/tag/v0.12.0

src/preload.ts
markypython: After updating deps, it was complaining about returning a non-void value in for each.  The alternative would be to use `.map` but then we are returning an array that is unused.

src/app.ts
markypython: I can switch it back to `import * from`.  The tsconfig update caused synthetic default imports to no longer work.

src/app.ts
timche: Interesting, thanks for sharing. Why would they change this to commonjs? Couldn't find any explanation.

package.json
```diff
@@ -22,36 +21,39 @@
...
+    "lint-staged": "^9.2.5",
     "npm-run-all": "^4.1.5",
     "prettier": "^1.16.4",
-    "stylelint": "^9.10.1",
+    "stylelint": "^10.1.0",
     "stylelint-config-prettier": "^5.0.0",
-    "stylelint-config-xo": "^0.13.0",
-    "typescript": "^3.3.4000",
+    "stylelint-config-xo": "^0.15.0",
+    "typescript": "~3.5.3",
```
timche: Why did this change from `^` to `~`?

src/app.ts
markypython: I assume that they changed it so it could be used by commonjs apps.

src/preload.ts
timche: We could use `for` loop instead.

package.json
markypython: It was trying to pull in TypeScript 3.6.x which one of the packages is not yet compatible with.

src/app.ts
timche: I think using `import * as electronContextMenu from 'electron-context-menu'` should work too then?

markypython: Yes, but it isn't the right solution according to this.

https://stackoverflow.com/a/35706271/5641674

timche: Thanks, we stick to this then 👍

src/preload.ts
markypython: After thinking about it, it would probably make more sense to leave it as a `.then` as we don't need to wait for each promise to resolve before attaching the next button listener.

PR #147, opened 2019-09-02
Auto update setting

### LINKED ISSUE #92: Make "Silent auto-updates" optional
First of all I understand that auto-update is extremely useful for 99% of the users (especially non-technical ones).  However in this day-and-age, especially for the security-conscious user, auto-update also implies that the tool you are using is downloading new code behind your back...  (This is why I consider it as a "bug" and not a "feature" request.)

Therefore I would strongly suggest to make this "silent auto-update" an optional feature, perhaps "opt-out" (i.e. enabled by default, but allowing one to disable it if needed).

Moreover I don't know how this auto-update would work on Linux where the package is installed in system-managed folders, and changing the installed files would require `root` access.

markypython:
Adds a setting to enable auto updating.  Also, this adds a menu item of "View logs" which will open the application logs to make debugging much easier.  Remembering where they are on different platforms and having to cd in the right directory is just a pain.

Fixes #92


timche:
Shouldn't we add a "Check for Updates..." additionally?

image

markypython:
@timche I have updated this PR to add the manual check for updates menu item

PR #158, opened 2019-09-18
Support starting app minimized to tray

cyfrost:
1. Uses the new `--start-minimized`  [command line switch](https://electronjs.org/docs/api/app#appcommandlinehasswitchswitch) to decide if the app should start minimized.
2. Bug fix: The tray context menu item "Show" wasn't being updated. Meaning, when the app is open, tray should say "Hide" and vice-versa. This has now been fixed.

Verified these changes on Linux and Windows, someone else has to dirty test it on macOS.

Thoughts?

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -30,12 +30,15 @@ initAutoUpdates()
 
 electronContextMenu({ showCopyImageAddress: true, showSaveImageAs: true })
 
+const shouldStartMinimized = app.commandLine.hasSwitch('start-minimized')
+
 app.setAppUserModelId('io.cheung.gmail-desktop')
 
 let mainWindow: BrowserWindow
 let replyToWindow: BrowserWindow
 let isQuitting = false
 let tray: Tray
+let trayContextMenu: Menu
 
 if (!app.requestSingleInstanceLock()) {
   app.quit()
@@ -67,7 +70,8 @@ function createWindow(): void {
       nodeIntegration: false,
       nativeWindowOpen: true,
       preload: path.join(__dirname, 'preload')
-    }
+    },
+    show: !shouldStartMinimized
   })
 
   if (lastWindowState.fullscreen && !mainWindow.isFullScreen()) {
@@ -92,6 +96,16 @@ function createWindow(): void {
       mainWindow.hide()
     }
   })
+
+  mainWindow.on('hide', () => toggleAppVisiblityTrayItem(false))
+
+  mainWindow.on('show', () => toggleAppVisiblityTrayItem(true))
+
+  function toggleAppVisiblityTrayItem(isMainWindowVisible: boolean) {
+    trayContextMenu.getMenuItemById('show-win').visible = !isMainWindowVisible
+    trayContextMenu.getMenuItemById('hide-win').visible = isMainWindowVisible
+    tray.setContextMenu(trayContextMenu)
+  }
 
   ipc.on('unread-count', (_: Event, unreadCount: number) => {
     if (is.macos) {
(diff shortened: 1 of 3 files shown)


markypython:
@cyfrost What are the use cases for starting the app minimized?  I'm assuming there are good use cases but I'm not familiar with what that might be.

cyfrost:
@markypython, My use-case is to have Gmail run at startup but minimized to the tray (to save it blasting open on my face on every boot). Other than that though, I can't think of anything else.

timche:
Thanks for this PR @cyfrost! As I find this feature quite useful as well, I'd like to have it configurable through the menu bar in `Settings` too. Could you implement that as well please?

cyfrost:
@timche, did you mean to have a "Auto start as minimized" option?

timche:
Correct.

src/app.ts
```diff
@@ -157,19 +171,31 @@ app.on('ready', () => {
     ]
 
     if (is.linux) {
-      contextMenuTemplate.unshift({
-        click: () => {
-          mainWindow.show()
+      contextMenuTemplate.unshift(
```
markypython: Could we keep this as a single item instead of two separate items?  Or would that get too messy?

cyfrost: That's a good idea indeed.

cyfrost: > Correct. More like `Start minimized` maybe?

@timche, I'm trying to understand: "Start minimized" can mean different things, for instance, when "Start minimized" is enabled and the user tries to launch Gmail from Launchpad, it could start silently in the tray, And the another meaning could be that "Start Minimized" will actually auto-start Gmail at system startup minimized to the tray. Which of this did you mean?

cyfrost:
> Correct. More like `Start minimized` maybe?

@timche, I'm trying to understand: "Start minimized" can mean different things, for instance, when "Start minimized" is enabled and the user tries to launch Gmail from Launchpad, it could start silently in the tray, And the another meaning could be that "Start Minimized" will actually auto-start Gmail at system startup minimized to the tray. Which of these did you mean?

cyfrost:
> Overall, this looks good. One thing I've been trying to do is move related functionality into separate files to better modularize the application. Could we move all of the tray logic to a separate file?
> 
> Thanks for the PR!

@markypython Can that refactor wait until this PR is merged after all concerns are addressed? Thanks!

timche:
You are right @cyfrost. Probably a better name could be `Launch minimized`? It shouldn't auto-start on boot.

cyfrost:
@timche, that has been added. We'll now have two ways to launch Gmail in minimized view. 1) append the `--start-minimized` switch when running via CLI (like `gmail --start-minimized`) and 2) Simply via the `Settings > Launch Minimized` option. Thoughts?

image

timche:
Sorry for the slow progress on my side. First, thanks for implementing this, can't wait to have this feature in!

Would it be possible to have the argument called `--launch-minimized` as well to have it aligned with how the setting is called?

Will test this on macOS.

cyfrost:
Heh, no worries. I understand people have a life outside GitHub too ;)

timche:
I'm getting this error when I try to start it on macOS:

image

cyfrost:
@timche My bad, sorry. The `trayContextMenu.getMenuItemById` was supposed to handle the change between "Show" and "Hide" in the Tray icon of Gmail, but I forgot to add ignore checks if the platform is macOS on which we do not create a Tray instance.

timche:
Released in [v2.13.0](https://github.com/timche/gmail-desktop/releases/tag/v2.13.0)

timche:
@all-contributors please add @cyfrost for code, maintenance

PR #159, opened 2019-10-01
fix issue #88 by adding option to hide Menu bar

### LINKED ISSUE #88: Change menu in Windows and Linux
For some reason the window menu-bar (i.e. the one containing `Gmail Desktop`, `Settings`, etc.) is visible all the time, and there is no option to hide it.

(I don't know if this is by design, or an issue with my window manager, namely `i3`.  I'm using OpenSUSE Tumbleweed, and manually installed the `2.6.1` from the Debian `.deb` by extracting it into a custom folder.)

cyfrost:
This fixes #88 by adding a configurable option to set the visibility of the Menu bar for those who want to change it.

When the Menu bar is hidden, we can hit the Alt key once to bring it up.

image

With Menu bar visible (current look):

image

With Menu bar hidden (new look):

image

Thoughts?

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -21,6 +21,7 @@ import { init as initDebug } from './debug'
 import { init as initDownloads } from './downloads'
 import { platform, getUrlAccountId } from './helpers'
 import menu from './menu'
+import { setAppMenuBarVisibility } from './utils'
 
 import electronContextMenu = require('electron-context-menu')
 
@@ -83,6 +84,8 @@ function createWindow(): void {
   if (lastWindowState.maximized && !mainWindow.isMaximized()) {
     mainWindow.maximize()
   }
+
+  setAppMenuBarVisibility()
 
   mainWindow.loadURL('https://mail.google.com')
 
--- a/src/app.ts
+++ b/src/app.ts
@@ -85,7 +85,9 @@ function createWindow(): void {
     mainWindow.maximize()
   }
 
-  setAppMenuBarVisibility()
+  if (is.linux || is.windows) {
+    setAppMenuBarVisibility()
+  }
 
   mainWindow.loadURL('https://mail.google.com')
 
--- a/src/menu.ts
+++ b/src/menu.ts
@@ -63,7 +63,7 @@ const createAppearanceMenuItem = ({
     }
 
     if (setMenuBarVisibility) {
-      setAppMenuBarVisibility()
+      setAppMenuBarVisibility(true)
     }
   }
 })
--- a/src/menu.ts
+++ b/src/menu.ts
@@ -14,6 +14,7 @@ interface AppearanceMenuItem {
   key: ConfigKey
   label: string
   restartDialogText?: string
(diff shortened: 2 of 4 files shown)


src/app.ts
```diff
@@ -84,6 +85,8 @@ function createWindow(): void {
     mainWindow.maximize()
   }
 
+  setAppMenuBarVisibility()
```
markypython: Is this something we can set on the window when creating it with the default value?  Then we can call this function later?

src/menu.ts
```diff
@@ -137,6 +137,15 @@ const applicationMenu: MenuItemConstructorOptions[] = [
           config.set(ConfigKey.LaunchMinimized, checked)
         }
       },
+      {
+        label: 'Hide Menu bar',
```
markypython: I think it would make sense to add this to the appearance menu.  @timche Thoughts?

src/utils.ts
```diff
@@ -1,9 +1,19 @@
 import { app, BrowserWindow, dialog } from 'electron'
+import config, { ConfigKey } from './config'
 
 export function getMainWindow(): BrowserWindow {
   return BrowserWindow.getAllWindows()[0]
 }
 
+export function setAppMenuBarVisibility(): void {
+  const mainWindow = getMainWindow()
+  if (mainWindow) {
```
markypython: When would this conditional be false?

cyfrost: I cannot think of any, my bad. It's a redundant check and will be removed, thanks for pointing it out.

src/menu.ts
cyfrost: You're right, I'll move it there then.

src/app.ts
cyfrost: > When the menu is turned off, how do you access it?

As mentioned in the OP, we can hit the Alt key once to bring up the Menu bar. And to keep users informed, this message below will be shown when the User hides the menu bar each time.

image

This will be shown each time the Menu bar is hidden by the user. Should we show this message only once?

cyfrost: > Is this something we can set on the window when creating it with the default value? Then we can call this function later?

Yes, we can set the `autoHideMenuBar` property while creating the window but a separate call to `mainWindow.setMenuBarVisibility(boolean)` is required to make the Alt key work. Without making this call, the Menu bar visibility will be set as per the config value but hitting the Alt key will not bring it up.

cyfrost:
> I noticed that this setting is available on all platforms, but it should only be configurable on Linux and Windows. The menu bar on macOS is integrated on system level on top. Otherwise I'm happy to with this feature.

Ah, thanks for pointing it out. I keep missing the consideration of macOS everytime, silly me! Let me fix this.

cyfrost:
@timche As discussed, [this](https://github.com/timche/gmail-desktop/pull/159/commits/426edad69e8df49b9a8e3d007fed776112c6f607) commit checks for macOS before allowing Menu bar visibility changes.

timche:
Thanks a lot! 👏

PR #164, opened 2019-10-07
Unread count for greater than 9 emails

markypython:
The regex for the unread count was only catching single number values.


timche:
Good find! 👏

Issue #36, opened 2019-03-13
Mac menu bar icon


mskelton:
It would nice to have a menu bar icon showing the count of unread emails in the Mac menu bar.  I envision something like Todoist (see screenshot below).

![2019-03-12_21-52-44](https://user-images.githubusercontent.com/25914066/54250292-322bc280-4511-11e9-8708-60e26486c97f.png)

timche:
I don't have a MB with touch bar, would be a cool addition though :) 👍

markypython:
My apologies, I meant the menu bar at the top of the screen.

timche:
Oh okay. I first thought of it, but I personally dislike it. I'd rather make it configurable and you can choose between dock, menu bar or dock + menu bar. What do you think?

markypython:
I definitely like the idea of making it configurable.

zzpxyx:
Do we have any plan on implementing this? I've done some quick experiments and it seems feasible?

timche:
I'm not sure why I closed this issue, but contributions are very welcome adding this feature!

zzpxyx:
No problem. I'm still learning here, but I'll see if I can come up with something. Thanks!

PR #165, opened 2019-10-16
Add Support for macOS Menu Bar Icon

zzpxyx:
This pull request adds support for the menu bar icon on macOS. This fixes Issue #36. A new settings menu item is also added so that the user can hide the system tray (menu bar) icon.

This pull request also fixes a null reference issue during application startup on Windows.


timche:
Great job! Works good for me. 

1. As per the issue description, we should display the number of unread messages? 
1. The icon could be monochrome like others: 
image. I'm not exactly sure how it could look like, but I can maybe come up with something.

src/menu.ts
```diff
@@ -151,6 +151,18 @@ const applicationMenu: MenuItemConstructorOptions[] = [
...
       },
+      {
+        label: is.macos ? 'Hide Menu Bar Icon' : 'Hide System Tray Icon',
+        type: 'checkbox',
+        checked: config.get(ConfigKey.HideTrayIcon),
+        click({ checked }: { checked: boolean }) {
+          config.set(ConfigKey.HideTrayIcon, checked)
+          showRestartDialog(
+            checked,
+            is.macos ? 'hiding menu bar icon' : 'hiding system tray icon'
```
markypython: This might read better by changing the setting to "Enable Menu Bar Icon" so then the dialog can say: "To disable the menu bar icon, please restart."

```suggestion
            is.macos ? 'the menu bar icon' : 'the system tray icon'
```

src/app.ts
```diff
@@ -34,6 +35,22 @@ electronContextMenu({ showCopyImageAddress: true, showSaveImageAs: true })
 const shouldStartMinimized =
   app.commandLine.hasSwitch('launch-minimized') ||
   config.get(ConfigKey.LaunchMinimized)
+const trayIcon = nativeImage.createFromPath(
```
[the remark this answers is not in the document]
markypython: Also, what is the purpose of the `nativeImage` stuff?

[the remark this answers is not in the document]
zzpxyx: Sure. I'll add a newline.

The `nativeImage` stuff is just being defensive as I'm not familiar with the low-level memory management in Electron. I was afraid that loading images from the disk all the time might affect performance or bring in memory leak. If we think that it won't be a big problem, I can change it back to use icon paths.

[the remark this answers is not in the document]
markypython: I'm fine with the `nativeImage` stuff, I was just asking since I'm not familiar with it 😄 

Thanks for the thoughts about performance!!

src/menu.ts
zzpxyx: OK, I'll make the code change. Thanks!

zzpxyx:
> Great job! Works good for me.
> 
>     1. As per the issue description, we should display the number of unread messages?
> 
>     2. The icon could be monochrome like others:
>        image I'm not exactly sure how it could look like, but I can maybe come up with something.
> 
>     3. It should probably an opt-in option and not shown by default.

Thanks for the review. Some responses:

1. Sure. I didn't know that it would be possible to show the number, but I think I've found something promising. Personally, I just need to know whether there are new emails or not, and I don't really care about the number as I'll open the inbox and read them all anyway. Maybe I can also make it configurable.
1. A monochrome icon would be nice. As mentioned above, I'd like to have a set of binary icons without the unread count. I'll also experiment on the icons and see what I get.
1. I left it on so that all platforms would have the same behavior. I guess it's fine to default to off on all platforms.

markypython:
> I left it on so that all platforms would have the same behavior. I guess it's fine to default to off on all platforms.

@timche Based on the fact that the tray shows up by default, I would agree with the decision to make it always on.

src/app.ts
```diff
@@ -35,6 +36,23 @@ const shouldStartMinimized =
   app.commandLine.hasSwitch('launch-minimized') ||
   config.get(ConfigKey.LaunchMinimized)
 
+const trayIcon = nativeImage.createFromPath(
```
markypython: I also just thought that creating the native image could be extracted to a helper function that takes a boolean of unread true/false as that is the only difference between the variables.

zzpxyx: Yeah, good point. I'll make the code change.

src/helpers.ts
```diff
@@ -12,3 +14,22 @@ export const platform = selectPlatform({
   linux: 'linux',
   windows: 'windows'
 })
+
+/**
+ * Create a tray icon.
+ *
+ * @param unread True to create the unead icon; false to create the normal icon.
+ */
+export function createTrayIcon(unread: boolean): nativeImage {
```
markypython: You probably want to use the interface here for the return type.

```suggestion
export function createTrayIcon(unread: boolean): NativeImage {
```

zzpxyx: OK. The confusing part is that `nativeImage` is a type alias or something of `NativeImage`, but I agree that using `NativeImage` seems to be the best here.

cyfrost:
Nice work, this PR also fixes #167

timche:
Would it be possible to split this PR to fix #167, since there is some work required for #36?

zzpxyx:
Thanks for splitting the fix!

After test-driving the development version for some days, I think the sweet spot is to omit the unread count when it is zero. In this case, we only need one monochrome icon. Now I no longer feel strongly to have binary icons. Maybe we can make it configurable if necessary in the future. To sum it up, the only missing part for this PR is a monochrome icon for the macOS menu bar. I'm trying to draw one, but I'll have to admit that I'm not talented in this, so it may take a while. Feel free to step in if you have an icon. :smile:

timche:
I have created a monochrome version of the logo. You can cherry-pick this commit: https://github.com/timche/gmail-desktop/commit/c1dfa0b207c61759c51b6390b1185ed1d79b0b47

timche:
I understand your point of having consistent behaviour across platforms, but I think it should be different on macOS since we have the app with unread count already in the dock and now also in the menu bar. On Windows though, there is no dock and therefore the tray bar is the only place we we can show unread messages to the user.

My proposal: Make menu bar icon optional with the setting to enable it + another option to show show only in menu bar, but hide in dock when you quit (cmd + q) for example. Some apps offers this behaviour like Airmail.

zzpxyx:
> I understand your point of having consistent behaviour across platforms, but I think it should be different on macOS since we have the app with unread count already in the dock and now also in the menu bar. That way it feels kinda cluttered. On Windows though, there is no dock and therefore the tray bar is the only place we can show unread messages to the user.
> 
> My proposal: Make menu bar icon optional with the setting to enable it + another option to show show only in menu bar, but hide in dock when you quit (cmd + q) for example. Some apps offers this behaviour like Airmail.

Thanks for the icon. Can you elaborate on the second option in your idea? My understanding is that we cannot turn off the menu bar icon and the dock icon at the same time. I'm a bit confused about what exactly the option does and how the option is related to quitting the application.

timche:
So I think there should be 3 options:

1. Dock only (Default)
1. Menu bar only
1. Dock + Menu bar

All my apps with unread badges are in the dock and having Gmail Desktop additionally in the menu bar feel a little too much for me.

zzpxyx:
OK, so there are multiple problems here.

1. The new menu bar icon only works in the dark theme. I have to change it to icon template so that it works in both light and dark themes. However, there is a bug in Electron 6 that makes the icon semi-transparent under the dark theme. I tried Electron 7 and the bug was fixed. Not sure if we want to upgrade now.
1. I also prototyped (dev branch on my fork) the idea of three-options in macOS icon settings. The biggest problem is about the menu bar only mode. After hiding the dock icon, the application menus are gone, including the settings menu. I couldn't find an easy way to switch to other icon options, so I got stuck in that mode. In addition, the application disappears in the CMD+TAB switcher. Both of these issues make this mode less desirable to use. We could tweak the context menu for the menu bar icon to make it work for sure, but that seems an overkill to me. Of course, I'm not very familiar with Electron and I may have missed some obvious workarounds.

If we stick to the previous two-option idea, I'm fine with hiding the menu bar icon by default if that's the concern, as hiding by default is recommended in Apple's [documentation](https://developer.apple.com/design/human-interface-guidelines/macos/extensions/menu-bar-extras/) anyway.

timche:
> The new menu bar icon only works in the dark theme. I have to change it to icon template so that it works in both light and dark themes. However, there is a bug in Electron 6 that makes the icon semi-transparent under the dark theme. I tried Electron 7 and the bug was fixed. Not sure if we want to upgrade now.

Could we provide two versions of the icon instead, white and black?

> If we stick to the previous two-option idea, I'm fine with hiding the menu bar icon by default if that's the concern, as hiding by default is recommended in Apple's documentation anyway.

Fine with it 👍 

Thanks a lot for the feedback!

zzpxyx:
> > The new menu bar icon only works in the dark theme. I have to change it to icon template so that it works in both light and dark themes. However, there is a bug in Electron 6 that makes the icon semi-transparent under the dark theme. I tried Electron 7 and the bug was fixed. Not sure if we want to upgrade now.
> 
> Could we provide two versions of the icon instead, white and black?
> 
> > If we stick to the previous two-option idea, I'm fine with hiding the menu bar icon by default if that's the concern, as hiding by default is recommended in Apple's documentation anyway.
> 
> Fine with it +1
> 
> Thanks a lot for the great feedback!

Theoretically yes, we could provide two versions of the icon, but I think that's an overkill as the bug in Electron has already been fixed upstream. My experiment showed that we would only need to change two lines when migrating to Electron 7, so it's not that hard. In terms of the bug itself, it makes the menu bar icon on the dark theme a bit gray, but the icon is still functioning well otherwise. To sum it up, I think the best option for now is to test and merge this PR first, and then I'll open another PR to bump the Electron version to 7.

timche:
Alright, let's do it like this 👍 

Is there anything left to do or should I test again?

zzpxyx:
> 
> 
> Alright, let's do it like this 👍
> 
> Is there anything left to do or should I test again?

I think it's done on my side. Feel free to test it again and see if I've missed anything. Thanks!

PR #169, opened 2019-10-19
update dependencies

cyfrost:
The output of `ncu` displaying current versions and latest available versions:

image


cyfrost:
Looks like `xo` being updated from v0.24.0 to v0.25.3 has created linting issues (most of which are `@typescript-eslint/strict-boolean-expressions`.

PR #172, opened 2019-11-24
Upgrade to Electron 7 to Solve Menu Bar Icon Issue

zzpxyx:
With Electron 6, the macOS menu bar icon displays as semi-transparent/grayish under the dark theme. This is an upstream issue with Electron and has been fixed under Electron 7. This PR upgrades the Electron dependency to version 7 and makes all necessary changes to make it build and run. Preliminary testing shows that the upgrade should be fine, but further testing may be needed.


timche:
Good job! Have tested on macOS, everything seems working. Thanks a lot!

timche:
@all-contributors please add @zzpxyx for code

Issue #79, opened 2019-04-13
Quick access from dock menu (macOS)


timche:
Inspired by: https://github.com/sindresorhus/caprine#quick-access-to-conversations-from-the-dock-menu-macos-only

PR #183, opened 2020-02-14
adding quick actions dock menu for macOS

jyoo980:
Hi!

I worked on Caprine, and I noticed that this app takes a lot of inspiration from it. I took a hand at implementing #79.  I created dock menu options for 

* `Compose`
* `Inbox`
* `Sent`

But I didn't create one for `All mail`, since I thought that was a bit redundant when we already have the `Inbox` option. Thanks!

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -23,6 +23,7 @@ import { platform, getUrlAccountId, createTrayIcon } from './helpers'
 import menu from './menu'
 import { setAppMenuBarVisibility, cleanURLFromGoogle } from './utils'
 import ensureOnline from './ensure-online'
+import { buildDockMenu } from './dock-menu'
 
 import electronContextMenu = require('electron-context-menu')
 
@@ -132,6 +133,10 @@ function createWindow(): void {
       }
     }
   })
+
+  if (is.macos) {
+    app.dock.setMenu(buildDockMenu(mainWindow))
+  }
 }
 
 function createMailto(url: string): void {
--- a/src/app.ts
+++ b/src/app.ts
@@ -2,7 +2,7 @@ import * as fs from 'fs'
 import * as path from 'path'
 import {
   app,
-  ipcMain as ipc,
+  ipcMain,
   shell,
   BrowserWindow,
   Menu,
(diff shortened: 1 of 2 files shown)


jyoo980:
@timche 

Thanks for the quick feedback. I'll take a look at doing that!

src/app.ts
```diff
@@ -137,6 +137,11 @@ function createWindow(): void {
   if (is.macos) {
     app.dock.setMenu(buildDockMenu(mainWindow))
   }
+
+  // TODO: create ipcMain.on(..) events for opening inbox,compose, and sent views
+  ipcMain.on('open-inbox-view', () => {
```
jyoo980: @timche 

I've started on the `ipc` method of doing this as you suggested, but I'm running into an issue where `ipc` here is always undefined. It looks like the import isn't working properly. I've ran a few searches but they all seemed to point me toward
 
* removing `node_modules` and reinstalling (tried that)
* not correctly importing it (followed the documentation)

I've also noticed that you use it in `preload.ts`, but I haven't been able to confirm whether it works there or not, since I don't think it's actually being run/used AFAIK.

Any thoughts?

src/dock-menu.ts
```diff
@@ -1,22 +1,25 @@
...
-): MenuItemConstructorOptions {
+function createInboxButton(): MenuItemConstructorOptions {
   return {
     label: 'Inbox',
-    async click() {
-      await mainWindow.loadURL(`${baseGmailUrl}/#inbox`)
-      mainWindow.show()
+    click() {
+      // TODO: investigate why ipc is undefined here
+      ipc.send('open-inbox-view')
```
jyoo980: @timche 

I've started on the `ipc` method of doing this as you suggested, but I'm running into an issue where `ipc` here is always undefined. It looks like the import isn't working properly. I've ran a few searches but they all seemed to point me toward
 
* removing `node_modules` and reinstalling (tried that)
* not correctly importing it (followed the documentation)

I've also noticed that you use it in `preload.ts`, but I haven't been able to confirm whether it works there or not, since I don't think it's actually being run/used AFAIK.

Any thoughts?

timche: 1. You are importing `ipcRenderer` which is only supposed to be imported in the renderer and not the main process.
2. To send a message from the main process to the renderer, you need use [BrowserWindow.webContents.send()](https://www.electronjs.org/docs/api/web-contents#contentssendchannel-arg1-arg2-).
3. We have a helper function in [`utils.ts`](https://github.com/timche/gmail-desktop/blob/master/src/utils.ts) called [`sendChannelToMainWindow()`](https://github.com/timche/gmail-desktop/blob/master/src/utils.ts#L23:L28) that you can use instead.

Issue #187, opened 2020-04-13
show destination URL for links hovered over


Viqsi:
One very useful way to avoid phishing is to be able to see the URL that a link is actually going to, Unfortunately, this isn't presently possible in Gmail Desktop. Most browsers handle this via a status bar, showing therein the destination URL of the link presently being hovered over. It would be nice if this, or some kind of tooltip, or some other display of the current hovered-over link's destination could be implemented.
Thanks!

timche:
We have added a confirmation dialog with showing the URL before the URL will be opened in the browser with #188. Hosts can be trusted for the future. It's similar to what VSCode has implemented.

PR #188, opened 2020-04-22
confirm destination URL before open link

hell0-Wor1d:
Show the real destination URL to avoid phishing attempt.

#187

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -319,9 +319,9 @@ app.on('before-quit', () => {
       .showMessageBox({
         type: 'info',
         buttons: ['Open Link', 'Cancel'],
-        title: `Do you want to open the external website "${cleanURL}" in your default browser?`,
+        message: `Do you want to open the external website "${cleanURL}" in your default browser?`,
         checkboxLabel: `Trust all links on ${origin}`,
-        message: cleanURL
+        detail: cleanURL
       })
       .then(({ response, checkboxChecked }) => {
         if (response !== 0) return
--- a/src/app.ts
+++ b/src/app.ts
@@ -7,7 +7,8 @@ import {
   BrowserWindow,
   Menu,
   Tray,
-  MenuItemConstructorOptions
+  MenuItemConstructorOptions,
+  dialog
 } from 'electron'
 import { is } from 'electron-util'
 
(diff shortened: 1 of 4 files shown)


src/app.ts
```diff
@@ -305,6 +306,12 @@ app.on('before-quit', () => {
       event.newGuest = win
     }
 
-    shell.openExternal(cleanURLFromGoogle(url))
+    dialog.showMessageBox({
+      type: 'info',
+      buttons: ['Yes', 'No'],
```
timche, line 311:
```suggestion
      buttons: ['Open Link', 'Cancel'],
```

src/app.ts
```diff
@@ -305,6 +306,30 @@ app.on('before-quit', () => {
...
+    }
+
+    dialog
+      .showMessageBox({
+        type: 'info',
+        buttons: ['Open Link', 'Cancel'],
+        title: `Do you want to open the external website "${cleanURL}" in your default browser?`,
+        checkboxLabel: `Trust all links on ${origin}`,
+        message: cleanURL
+      })
```
timche, line 325:
```suggestion
    dialog
      .showMessageBox({
        type: 'info',
        buttons: ['Open Link', 'Cancel'],
        message: `Do you want to open the external website "${cleanURL}" in your default browser?`,
        checkboxLabel: `Trust all links on ${origin}`,
        detail: cleanURL
      })
```

timche:
There are also other places where external URLs are opened:
- https://github.com/timche/gmail-desktop/blob/master/src/app.ts#L301
- https://github.com/timche/gmail-desktop/blob/master/src/app.ts#L308

Could you please wrap the stuff in a function and re-use that in these places too?

hell0-Wor1d:
Done !
It look like this on Windows : 
Capture

src/app.ts
```diff
@@ -168,6 +169,10 @@ function addCustomCSS(windowElement: BrowserWindow): void {
   }
 }
 
+function openExternalUrl(url: string): void {
+  shell.openExternal(url)
+}
+
```
timche, line 175: 😂 

You understood me wrong.

I meant you should wrap L313-L337 into a function, so everywhere where an external URL can be opened, should have the same behavior with trusting URLs.

timche:
Awesome, cheers for your work 🙇

timche:
@all-contributors please add @hell0-Wor1d for code

Issue #204, opened 2020-07-28
Cannot edit config file on Linux


ipkpjersi:
I am presented with this error when attempting to edit the config file:
![](https://i.imgur.com/fkanp27.png)
- Gmail Desktop Version: 2.20.2
- OS Version: Ubuntu 18.04 with Xfce 4.12

Steps to Reproduce:

1. Install via deb on Ubuntu 18.04 File -> Open 
2. Go to Settings -> Advanced -> Edit Config File

timche:
This has been introduced with the upgrade of Electron: ed373173d706214c89b7386c7501d292283ef390

We need to upgrade `electron-store` that supports the latest Electron version.

pilotstew:
Any workaround for this?

PR #215, opened 2020-11-02
Update dependencies

mskelton:
I just updated my system to the macOS Big Sur beta and the older version of Electron doesn't match the window style of Big Sur.  This will fix that by updating to the latest Electron version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,16 +20,16 @@
     "dist": "tsc && electron-builder --macos --linux --windows --publish never"
   },
   "dependencies": {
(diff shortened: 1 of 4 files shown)


timche:
This should also fix #204. Thanks a lot!

mskelton:
@timche Looks like you just commented, but I'm assuming you are fine with me merging right?

timche:
I was waiting for the build to complete so I can test it.

timche (approved):
LGTM

PR #229, opened 2021-03-10
Fix taskbar icon on linux

wennaspeedy:
I fixed taskbar icon in linux.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ yarn-error.log
 dist/
 dist-js/
 *.tgz
+package-lock.json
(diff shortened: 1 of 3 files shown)


timche:
Can you please rebase your branch against `master`? Your PR should build successfully then and produce a test build for all OS that are available as artefacts in the GitHub Action.

timche:
Thanks for providing the fix. I've narrowed it down to `icon: path.join(__dirname, '../build/icon.png')` and set the value only when it's Linux here: https://github.com/timche/gmail-desktop/pull/228/files#diff-841254fe75488c1bd4cd7f68f00b4be0e48dcfbc4a16b45847b68295e0e3b27b

wennaspeedy:
You're welcome, I like this app.

PR #287, opened 2021-06-23
Fix 'hide menu bar' menu item

alecmev:
Leaves empty space on top once disabled, and after restarting the app, once activated with Alt, fails to render, can only see the outline around the items. Once one of the items is selected, though, the submenu renders correctly. But this is a separate issue.


timche:
Thanks a lot! :bow:

timche (approved):

PR #288, opened 2021-06-23
Handle mailto in argv, support full syntax

alecmev:
To make it work on Linux and, [I assume](https://stackoverflow.com/a/38205984/242684), Windows.

### DIFF
--- a/src/main/app.ts
+++ b/src/main/app.ts
@@ -1,7 +1,7 @@
 import { app } from 'electron'
 import config, { ConfigKey } from './config'
+import { handleMailto } from './mailto'
 import { getMainWindow, showMainWindow } from './main-window'
-import { sendToSelectedAccountView } from './account-views'
 import { appId } from '../constants'
 
 let isQuittingApp = false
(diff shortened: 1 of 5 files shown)


src/main/app.ts
```diff
@@ -35,20 +38,24 @@ export async function initApp() {
...
       if (mainWindow.isMinimized()) {
         mainWindow.restore()
       }
 
       mainWindow.show()
+
+      const mailto = argv[argv.length - 1]
+      if (mailto.startsWith('mailto:')) {
+        handleMailto(mailto)
+      }
```
timche, line 53: Since `argv` is only handled when trying to launch a second instance, what if the launch is the first instance?

alecmev: Good catch, addressed :wink:

alecmev:
@timche I've fixed the rebase, but I wasn't sure how to merge the changes in `src/main/app.ts`. Can you, please, take a look? I'm unable to do more right now, I'm sorry. Thank you!

alecmev:
@timche Hi! What's the status of this? Do you need help?

alecmev:
@timche Hi! Any chance of revisiting this? Would love to get this merged.

PR #301, opened 2022-02-07
Fix missing application icon on Linux

buckmelanoma:
When running gmail-desktop on Linux/X11 a generic fallback icon is shown in the titlebar.  The path for the included icon appears to be slightly off.  This PR changes the path to fix that problem.

Before:
before

After:
after


timche:
Thanks for fixing this!

timche:
Fixes #278

PR #302, opened 2022-02-09
fix some yarn scripts on windows

DinCahill:
These commands need double quotes to run on Windows. Works on Mac and Linux too.

Also removed a `--`

> yarn run v1.22.17
> $ yarn prettier -- "**/*.{ts,tsx,js,css,json}"
> warning From Yarn 1.0 onwards, scripts don't require "--" for options to be forwarded. In a future version, any explicit "--" will be forwarded as-is to the scripts.


DinCahill:
Windows CI failed due to a network problem 😅

timche (approved):

PR #303, opened 2022-02-09
fix label icons

DinCahill:
Label icons on the left bar show up as boxes.

For some reason, the css `mask` property is showing as invalid, even in newer versions of electron.

I copied the broken css selector from its source in the dev tools, changed `mask` to `-webkit-mask`, and added it to the custom stylesheet.

The labels now show up correctly in light mode.

Dark mode is a bit funky. The purple label is a slightly lighter grey than the other default coloured labels. The correct behaviour would be to leave the coloured labels untouched.

Before:

Labels Before

After:

Labels After

CSS:

Labels CSS

### DIFF
--- a/src/main/account-views/preload/dark-mode.ts
+++ b/src/main/account-views/preload/dark-mode.ts
@@ -51,7 +51,9 @@ function enableDarkMode(): void {
         // Hangouts Conversations
         '.aj2',
         // Hangouts Phone
-        '.a8V'
+        '.a8V',
+        /* Labels */
+        '.aEe, .aEc.aHS-bnr .qj'
       ]
     }
   )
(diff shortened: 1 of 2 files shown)


src/main/account-views/preload/dark-mode.ts
```diff
@@ -51,7 +51,9 @@ function enableDarkMode(): void {
         // Hangouts Conversations
         '.aj2',
         // Hangouts Phone
-        '.a8V'
+        '.a8V',
+        /* Labels */
```
timche, line 55:
```suggestion
        // Labels
```

timche (changes requested):
That's awesome! Thanks a lot.

timche (approved):

PR #307, opened 2022-02-11
hide google one account element

timche:


timche:
@duong-le can you please test the build from the artifacts [here](https://github.com/timche/gmail-desktop/actions/runs/1830735954)? The Google One ring should be hidden (if that's fine).

duong-le:
@timche it worked! But why don't you resize the ring so that it fits the avatar well?

duong-le:
@timche I resized the ring by setting the width and height of the SVG to 26px but was unable to move it to the right. If you need a quick Meet call to find the solution, I'm ok with that
image

PR #308, opened 2022-02-12
fix google chat status in compact header mode

DinCahill:
Here it is:
Screenshots of fixed compact header bar


timche:
Great work, thanks!

PR #310, opened 2022-02-12
open main window on some actions (macOS)

DinCahill:
On Mac, you can press Cmd+W to close the window, and still access the menu bar. Some options which I expected to reopen the main window didn't do so, but others did.

Now, the main window will open/focus when you press any of: Gmail Settings, Add Account, Edit Account, Remove Account.

It will also open/focus when handling a mailto link.


timche:
Nice catch, thanks!

PR #313, opened 2022-02-13
Fix/move google chat status menu

DinCahill:
The position of the Google Chat status dropdown is calculated in javascript. When compact header mode is enabled, it isn't positioned correctly, presumably because of the elements we've hidden.

They position it with the `top` and `left` properties, so I've used `transform: translateX(-100px)` to apply a constant offset.

Before:
before narrow
before wide
After:
fixed narrow
fixed wide
And when `<body dir='rtl'>`:
arabic

### DIFF
--- a/src/main/account-views/custom-styles/style.css
+++ b/src/main/account-views/custom-styles/style.css
@@ -213,8 +213,12 @@
 }
 
 /* Header: Google Chat status dropdown menu */
-.compactHeader .tP {
+body:not([dir='rtl']).compactHeader .tP {
   transform: translateX(-100px);
+}
+
+body[dir='rtl'].compactHeader .tP {
+  transform: translateX(100px);
 }
 
 /* Header: Google One account ring */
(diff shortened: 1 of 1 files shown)


DinCahill:
Something like this?
```css
:not([dir='rtl']).gmail-desktop_compact-header
  [style*='top:'][style*='left:'][style*='visibility:']
```

timche:
As you're only targeting style values, I'm not 100% if this is unique enough. Are there any other attributes or paths you could use?

DinCahill:
> As you're only targeting style values, I'm not 100% sure if this is unique enough. Are there any other attributes or paths you could use?

There are no other attributes. It has a single hardcoded Gmail class, so if we used that, the selector could just be `:not([dir='rtl']).gmail-desktop_compact-header .tP`.

timche:
Thanks for explaining. 

Let's use the Gmail class instead, although it's not good, but safer, because `[style*='top:'][style*='left:'][style*='visibility:']` could also target another element with the same attributes.

timche (approved):

PR #316, opened 2022-02-18
use hide window buttons on macos only

DinCahill:
I don't think the "Hide", "Hide Others" and "Show All" buttons do anything on Windows or Linux.

I removed the "Hide" and "Hide Others" accelerators because they are provided by default on macOS. "Hide Others" now gets set to `Command+Option+H` instead of `Command+Shift+H`. (Safari and Chrome use `Command+Shift+H` to go Home, so we could potentially use that hotkey to go to the inbox or something.)

The `visible` property doesn't hide separators (electron/electron#3494), so I've used a quick and dirty solution.

### DIFF
--- a/src/main/menus/app.ts
+++ b/src/main/menus/app.ts
@@ -50,6 +50,24 @@ interface AppearanceMenuItem {
 }
 
 export function getAppMenu() {
+  const macOSWindowItems: MenuItemConstructorOptions[] = [
+    {
+      label: `Hide ${app.name}`,
+      role: 'hide'
+    },
+    {
+      label: 'Hide Others',
+      role: 'hideOthers'
+    },
+    {
+      label: 'Show All',
+      role: 'unhide'
+    },
+    {
+      type: 'separator'
+    }
+  ]
+
   const appearanceMenuItems: AppearanceMenuItem[] = [
     {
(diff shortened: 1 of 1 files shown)


src/main/menus/app.ts
```diff
@@ -472,21 +472,27 @@ export function getAppMenu() {
...
-          type: 'separator'
-        },
+        is.macos
+          ? {
+              type: 'separator'
+            }
+          : {
+              label: 'hidden separator',
+              visible: false
+            },
```
timche, line 495: This could be simplified like this:

```ts
        ...(is.macos
          ? ([
              {
                label: `Hide ${app.name}`,
                role: 'hide'
              },
              {
                label: 'Hide Others',
                role: 'hideOthers'
              },
              {
                label: 'Show All',
                role: 'unhide'
              },
              {
                type: 'separator'
              }
            ] as MenuItemConstructorOptions[])
          : []),
```

timche: I think you're missing the spread `...` operator. Look closely at my code example at the beginning.

DinCahill: Ah, I see, thanks!

timche (approved):
👍

Issue #328, opened 2022-04-10
Notification/Chat count for Gmail Workspace


enzorevo:
**OS**
macOS, Catalina 10.15.7

**Gmail Desktop Version**
3.0.0.alpha34

**Describe the bug**
The notification number, both in the dock and in the app panels, work only for classic GMail visualization. I believe this depends on new selectors for the new Google Workspace Gmail visualization, which has the Google Chat panels in the leftmost panel in addition. That would be great to have a notification count for both GChat messages + Mails, in that case!

Amazing work, anyways!

PR #340, opened 2022-09-07
fix: Unread message indicator in status bar

BapRx:
Update the CSS selector to fix the `getUnreadCount()` function.


BapRx:
This PR should fix #328

timche:
We're not accepting any PRs for v2 anymore, sorry!

PR #341, opened 2022-09-07
fix: Unread message indicator in status bar (develop)

### LINKED ISSUE #328: Notification/Chat count for Gmail Workspace

BapRx:
I just realized that the fix I did was on the main branch :sweat:, here's the same fix, tested on alpha build.

Closes #328

### DIFF
--- a/src/main/account-views/preload/gmail.ts
+++ b/src/main/account-views/preload/gmail.ts
@@ -22,7 +22,7 @@ const mailActions = {
   markAsSpam: 'sp'
 }
 
-const inboxElementSelector = 'div[role=navigation] a[href*="#inbox"]'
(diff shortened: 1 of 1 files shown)


timche:
Thanks a lot! :)

Issue #364, opened 2023-03-22
The "download completed" notification always expects the file to be in the Downloads folder


DonaldDuck313:
**OS**
Windows

**Gmail Desktop Version**
2.25.3.0

**Describe the bug**
When I download an attachment into a folder other than the Downloads folder, I get a notification saying "Download completed", but when I click on it it says that the file "C:\Users\myname\Downloads\file.pdf" doesn't exist. Obviously it doesn't exist, I put the file somewhere else. It should look in the folder where I put it instead of the Downloads folder.

**Steps to reproduce the behavior**

1. Find an email with an attachment.
2. Download the attachment.
3. Save the attachment in a folder other than the Downloads folder.
4. When you get a notification saying "Download completed", click on it.
5. You will get an error message because it looked in the wrong folder.

**Expected behavior**
Clicking on the notification should open the correct file instead of showing an error because it attempts to open the wrong file.

**Media**
![image](https://user-images.githubusercontent.com/40611243/226946692-377ee705-bd3f-4bec-ad18-1af3e0b4c689.png)

![image](https://user-images.githubusercontent.com/40611243/226946975-2323a99c-0d9c-4499-9dbf-6ccb19f7a018.png)

![image](https://user-images.githubusercontent.com/40611243/226947743-24196de2-76f2-4319-9abe-bc09caf04256.png)

It looks in the Downloads folder even though that's not where I saved the file. In this case it should try to open C:\Users\myname\AppData\Local\Temp\709673.pdf instead.

![image](https://user-images.githubusercontent.com/40611243/226948368-b8598d9f-3752-41b4-94d2-719ad109d658.png)

PR #376, opened 2023-07-06
downloads: use actual save path when opening

brendanmeyer:
Fix for #364 

using getSavePath to retrieve the downloaded files full path


timche:
Thanks for fixing this!

PR #489, opened 2026-01-08
Activate first account with unread mail on tray icon click

webcoyote:
## Summary
- When clicking the menubar icon, automatically switch to the first account that has unread emails before showing the window
- If multiple accounts have unread mail, activates the first one in the account list
- If no accounts have unread mail, shows the window normally without changing the selected account

## Test plan
- [x] Configure multiple accounts
- [x] Receive emails in a non-selected account
- [x] Click the menubar icon and verify it switches to the account with unread mail
- [x] Verify that if no accounts have unread mail, clicking the icon shows the currently selected account

### DIFF
--- a/packages/app/accounts.ts
+++ b/packages/app/accounts.ts
@@ -310,6 +310,20 @@ class Accounts {
 			0,
 		);
 	}
+
+	getFirstAccountWithUnread() {
+		for (const accountConfig of this.getAccountConfigs()) {
+			const instance = this.instances.get(accountConfig.id);
+
+			if (instance) {
+				const unreadCount = instance.gmail.store.getState().unreadCount;
+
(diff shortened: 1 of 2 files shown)


packages/app/accounts.ts
```diff
@@ -310,6 +310,22 @@ class Accounts {
...
+			if (instance) {
+				const unreadCount = instance.gmail.store.getState().unreadCount;
+
+				if (typeof unreadCount === "number" && unreadCount > 0) {
+					return accountConfig;
+				}
+			}
+		}
+
+		return undefined;
```
timche: `return undefined` is redundant and can be removed.

webcoyote: Fixed!

PR #498, opened 2026-03-14
Fix Linux tray

BlueManCZ:
- Fix tray icon left-click doing nothing on Linux - was using `setContextMenu` which intercepted left-click instead of firing the `click` event. Now uses both `setContextMenu` (for right-click menu) and a `click` handler (to toggle window visibility), which works on Electron v22+ thanks to the native D-Bus StatusNotifierItem implementation
- Remove platform-specific branching for tray setup - the `setContextMenu` + `click` handler pattern should work consistently across macOS, Windows, and Linux
- Fix tray context menu Show/Hide items using static `shouldLaunchMinimized` value instead of actual `window.isVisible()` state
- Fix tray context menu not updating after account rename or unread count change - rebuild and re-set the context menu on the tray when accounts or visibility change

## Test plan
- Left-click tray icon when window is visible — should hide the window
- Left-click tray icon when window is hidden — should show the window
- Right-click tray icon — should show context menu with accounts, Show/Hide, and Quit
- Hide window via tray, right-click tray — "Show" should be visible, "Hide" should be hidden
- Rename an account — tray context menu should reflect the new name without restart
- Verify on macOS and Windows as well

### DIFF
--- a/packages/app/tray.ts
+++ b/packages/app/tray.ts
@@ -8,7 +8,7 @@ import { main } from "@/main";
 export class AppTray {
   private tray: Electron.Tray | undefined;
 
-  private menu: Electron.Menu | undefined;
+  private contextMenu: Electron.Menu | undefined;
 
   private icon: Electron.NativeImage | undefined;
   private iconUnread: Electron.NativeImage | undefined;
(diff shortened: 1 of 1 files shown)


BlueManCZ:
https://github.com/user-attachments/assets/3f234b72-7158-484c-9a9d-5873c2e52be1

BlueManCZ:
Yes, I noticed similar behavior on my Gnome. But then I realized single left-click doesn't work for any application but that double-click works. I guess it depends how tray manager handles this and it can be slightly different between environments.

PR #500, opened 2026-03-14
Hide titlebar overlay on Linux when gtk-decoration-layout has no window controls

BlueManCZ:
Hides the `titleBarOverlay` window control buttons (minimize, maximize, close) on Linux when the user's `gtk-decoration-layout` setting indicates no window controls should be shown.

The detection reads `gtk-decoration-layout` - the standard mechanism GTK CSD applications use to determine which window buttons to render:

1. **GSettings** (`org.gnome.desktop.wm.preferences button-layout`) - the authoritative runtime source used by most desktop environments
2. **GTK settings files** (`~/.config/gtk-{3,4}.0/settings.ini`, `/etc/gtk-{3,4}.0/settings.ini`) - fallback for environments without GSettings

If the layout contains no `close`, `minimize`, or `maximize` entries, the titlebar overlay is disabled.

This is useful for tiling window managers (i3, sway, Hyprland, etc.) where window controls are redundant. Users can configure this with:

```sh
gsettings set org.gnome.desktop.wm.preferences button-layout ':'
```

### DIFF
--- a/packages/app/lib/linux.ts
+++ b/packages/app/lib/linux.ts
@@ -1,6 +1,6 @@
 import { platform } from "@electron-toolkit/utils";
 import * as childProcess from "node:child_process";
-import { readFileSync } from "node:fs";
+import { readFile } from "node:fs/promises";
 import { homedir } from "node:os";
 import { join } from "node:path";
 import { promisify } from "node:util";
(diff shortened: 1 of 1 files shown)


BlueManCZ:
Ideally yes - Electron's `titleBarOverlay` should natively respect `gtk-decoration-layout` from GSettings. Currently it's all-or-nothing: it either shows all three buttons or none, ignoring the user's decoration layout preference entirely. This is tracked in [electron/electron#44540](https://github.com/electron/electron/issues/44540).

That said, getting a fix upstream into Electron can take a while, and this is a small, self-contained workaround that can be removed once Electron handles it properly.

PR #730, opened 2026-08-08
add an option to position the workspace apps launcher on the right

timche:
The Workspace Apps launcher is fixed to the left side of the titlebar. Before it was grouped with the Unified Inbox button, it lived in the right-hand cluster, which some users prefer.

Stacked on #729 (Launcher Display) — review that one first.

## Changes

- New config key `workspaceApps.launcherPosition` (`"left"` | `"right"`), defaulting to `"left"` so existing behavior is unchanged.
- New **Launcher Position** select in Settings → Workspace Apps, below Launcher Display.
- `app-titlebar.tsx`: the launcher element is built once and rendered either in the existing left button group (next to Unified Inbox) or in the right-hand cluster, immediately before Saved Searches — so it sits next to Saved Searches when those exist and next to the downloads button when they don't. Its dropdown flips to `side="left"` on the right so it opens inward.
- Only the launcher moves. The Unified Inbox button stays on the left, and the left button group no longer renders at all when the launcher is on the right and Unified Inbox is hidden.

Both display modes (menu and inline) work in either position.

## Test plan

1. With a valid license and apps in Launcher Apps, confirm the launcher renders on the left by default, grouped with the Unified Inbox button.
2. Set **Launcher Position** to *Right* → the launcher moves into the right cluster, immediately left of the Saved Searches button.
3. With no saved searches configured → on Right, the launcher sits immediately left of the downloads button instead.
4. On Right with **Launcher Display** = *Menu*, open the menu → it opens toward the left (inward), not clipped at the window edge.
5. On Right with **Launcher Display** = *Inline* → all app icons render in the right cluster, spaced like the other titlebar icon buttons, and clicking/middle-clicking behaves as on the left.
6. On Right with Unified Inbox enabled and 2+ accounts → the Unified Inbox button stays on the left; no empty group or stray gap is left where the launcher was.
7. On Right with Unified Inbox disabled (or a single account) → the left group disappears entirely; navigation controls and account buttons keep their spacing.
8. Switch to the unified inbox → the launcher is disabled in both positions.
9. Set Right, restart the app → the setting persists.

### DIFF
--- a/packages/app/config.ts
+++ b/packages/app/config.ts
@@ -96,6 +96,7 @@ export const config = new Store<Config>({
     "workspaceApps.openBehavior": "tab",
     "workspaceApps.launcherApps": [],
     "workspaceApps.launcherDisplay": "menu",
+    "workspaceApps.launcherPosition": "left",
     "workspaceApps.showAccountColor": true,
     "workspaceApps.showAccountLabel": true,
     "workspaceApps.persistZoom": true,
--- a/packages/app/config.ts
+++ b/packages/app/config.ts
@@ -95,6 +95,7 @@ export const config = new Store<Config>({
     "workspaceApps.openInAppExcludedApps": [],
     "workspaceApps.openBehavior": "tab",
     "workspaceApps.launcherApps": [],
+    "workspaceApps.launcherDisplay": "menu",
     "workspaceApps.showAccountColor": true,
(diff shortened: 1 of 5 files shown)


timche:
Superseded: the launcher position is fixed on the right and no longer configurable, which #729 now does directly.

PR #793, opened 2026-08-12
add todo: html fullscreen leaves the app shell visible

timche:
Notes a bug found while testing #723: presenting in Slides and going fullscreen in Meet leave the titlebar and vertical tabs strip on screen, because nothing listens for `enter-html-full-screen` on a view's `webContents`.

Notes only — no behaviour change.


timche:
Folded into #723 — the note belongs with the change that made fullscreen reachable.

PR #806, opened 2026-08-14
make Docs' Edit menu cut, copy and paste work

timche:
Google Docs' own Edit menu offers to install the "Docs Offline" Chrome extension instead of cutting, copying or pasting. Two things are missing in Meru, both addressed here:

- The extension's `page_embed_script.js` (339 bytes, fetched from the Web Store CRX) only sets `window._docs_chrome_extension_exists` and friends — a marker for the clipboard permissions the extension's `content_capabilities` grant the page. The workspace app preload now sets that marker on `docs.google.com` in the main world, with only the two clipboard permissions listed so Docs doesn't take it as an offer of offline storage.
- The clipboard access itself: Docs cuts, copies and pastes through `document.execCommand`, which Electron gates behind the `enableDeprecatedPaste` web preference (off by default) plus the `deprecated-sync-clipboard-read` permission. Workspace app views now set the preference; the session's permission check handler already grants the permission.

Loading the real extension via `session.loadExtension` was the other route and is a dead end: Electron implements neither `content_capabilities` — the mechanism that grants Docs its clipboard access in Chrome — nor `web_accessible_resources`, and it would mean shipping Google's 135 KB service worker.

Risks: I can't run the app against real Google Docs from here, so none of this is verified end to end — in particular whether today's Docs is happy with the two globals alone (Wavebox shipped exactly these two, but years ago) or also wants `_docs_chrome_extension_features_version`. If it isn't, the menu keeps offering the extension, as it does today. Claiming the extension exists may also make Docs surface offline options that can't work, since there is no `chrome.runtime` to talk to. `enableDeprecatedPaste` is deprecated in Electron and will eventually be removed, which puts the paste back where it is today.

Test plan:
1. Open a document in Docs in-app, select some text, and use Edit → Copy, then Edit → Paste — both should act on the clipboard instead of showing the "install the Docs Offline extension" prompt.
2. Paste something copied from outside Meru through Edit → Paste; it should land in the document.
3. In the same document, check the File menu's offline entry — Docs now believes an extension is installed, so watch for an offline toggle that errors or hangs.

### DIFF
--- a/packages/app/workspace-app.ts
+++ b/packages/app/workspace-app.ts
@@ -5,7 +5,6 @@ import type { AccountConfig } from "@meru/shared/schemas";
 import { clamp } from "@meru/shared/utils";
 import {
   type SupportedWorkspaceApp,
-  WORKSPACE_APP_PRELOAD_ARGUMENTS,
   type WorkspaceAppBookmarkState,
   type WorkspaceAppOpenBehavior,
   workspaceApps,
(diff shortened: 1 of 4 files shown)


packages/app/workspace-app.ts
```diff
@@ -541,7 +541,15 @@ export class WorkspaceApp {
...
       session: this.account.instance.session,
       preload: getPreloadPath("workspace-app"),
-      viewOptions: options,
+      viewOptions: {
+        ...options,
+        webPreferences: {
+          ...options?.webPreferences,
+          // Docs cuts, copies and pastes from its Edit menu through
+          // `document.execCommand`, which Electron gates behind this preference
+          enableDeprecatedPaste: true,
```
timche: only enable if app is docs?

timche: Done in f53e00e. `enableDeprecatedPaste` is now `this.app === "docs"` at view creation, and the preload only injects the extension globals when the view carries a matching `--meru-docs-menu-clipboard` argument (same pattern as `GMAIL_PRELOAD_ARGUMENTS`). The pairing matters because `this.app` can be reassigned on in-place navigation while webPreferences are fixed at creation: a non-Docs view navigated to Docs would otherwise advertise paste support it doesn't have and fail silently — with the gate on both sides it shows today's install prompt instead.

timche: Reverted in 883b83d — back to enabling it for all workspace app views.

PR #824, opened 2026-08-16
run native messaging hosts for extension contexts

timche:
- Electron's built-in `connectNative` is a categorical refusal, not a discovery problem: `ElectronMessagingDelegate::IsNativeMessagingHostAllowed` returns `DISALLOW` unconditionally (verified in Electron 43 source and empirically against 7 manifest locations), so the facade replaces `runtime.connectNative`/`sendNativeMessage` on both `chrome` and `browser`.
- The bridge to main is a per-session privileged custom scheme (`protocol.handle`): extension contexts fetch it, the connect response streams framed messages back. `allowServiceWorkers` is required or SW fetches die before the handler; extension-page CSP gets `connect-src` extended at derive time. Since `protocol.handle` receives no `Origin` and any page in the session can reach the scheme, calls carry a per-extension bearer token regenerated each launch and written into that extension's derived facade copy — not web-reachable for 1Password (its `web_accessible_resources` doesn't cover the facade file).
- Host side matches Chrome: per-OS manifest discovery (Linux/macOS dirs for Chrome/Chromium/Edge/Brave/Vivaldi, Windows via `reg query`), `allowed_origins` enforcement, one host process per port with origin as argv[1], 4-byte-LE framing with the 1MB inbound cap, Chrome's error strings verbatim. Port disconnect, oversize, session teardown, and process exit all kill host processes (verified by PID).
- From real-machine (macOS) testing: Chromium never re-fetches an extension SW's script once registered, so later launches ran a stale facade with a dead token → every call 403'd. The loader now drops the extension's SW registration before each load (`clearStorageData` scoped to the extension's origin; id derived from `manifest.key`). Without this, no facade change ever reached an existing profile's SW.
- Also from that testing: derived copies drop `webRequest`/`webRequestAuthProvider` from permissions (inert in Electron; removes the renderer's `NOTREACHED "webRequest" not found` bursts entirely), and Gmail's `loadURL` rejections are caught and logged (`Failed to load URL`) instead of surfacing unhandled.
- The blank-Gmail `ERR_FAILED` is root-caused and fixed: `loadExtension` completing while Gmail's navigation was still provisional made Chromium abort that navigation (both failing runs died ~90ms after `Loaded extension`; successful runs had committed before it). Gmail's first navigation now waits for the session's extension loads (a failed extension setup logs and never blocks Gmail), and the initial load retries once if it still never commits.
- Diagnostics for any future empty view stay in: `render-process-gone` / `did-fail-load` / `did-fail-provisional-load` are logged for the Gmail view, and `MERU_EXTENSIONS_STRIP=<manifest keys>` (dev-only) derives extensions without the listed keys for bisecting.

Not verified: the race fix and retry against a real failing launch (the race is timing-dependent — several clean starts are the signal); the real 1Password host (it verifies the calling browser's code signature — a dev build may be refused until Add Browser sees a signed build); Touch ID; macOS/Windows manifest lookup paths (unit-tested as data only). Onboarding UI for Add Browser is a later slice.

Test plan (needs a machine with the 1Password desktop app):
1. `bun run extensions:download`, then `bun run dev` repeatedly (5+ starts) on an existing profile → Gmail renders every time; a `Retrying Gmail load` line, if it appears, means the abort happened and the retry covered it.
2. Log shows no 403s; 1Password's desktop-app probe either connects or fails on the host's own trust check, not on the bridge.
3. 1Password → Settings → Browser → Add Browser → watch for `Connected native messaging host` in Meru's log and 1Password reporting the desktop-app link (may require a signed build).
4. Quit Meru with the extension connected → no orphaned `1Password-BrowserSupport` process.

### DIFF
--- a/packages/app/account.ts
+++ b/packages/app/account.ts
@@ -4,10 +4,12 @@ import type { AccountConfig } from "@meru/shared/schemas";
 import type { VerticalTabsSessionWidth } from "@meru/shared/tabs";
 import type { SelectedDesktopSource } from "@meru/shared/types";
 import { app, type IpcMainEvent, ipcMain, type Session, session } from "electron";
+import { serializeError } from "serialize-error";
 import { blocker } from "./blocker";
 import { config } from "./config";
 import { extensions } from "./extensions";
 import { Gmail } from "./gmail";
+import { log } from "./lib/log";
 import { createBrowserWindow, getPreloadPath, loadRenderer } from "./lib/window";
 import { licenseKey } from "./license-key";
 import { main } from "./main";
@@ -46,7 +48,9 @@ export class Account {
 
     blocker.setupSession(this.session);
 
-    extensions.setupSession(this.session);
+    const extensionsLoaded = extensions.setupSession(this.session).catch((error: unknown) => {
+      log.error("Failed to set up extensions", { error: serializeError(error) });
+    });
 
     this.setSpellCheckerLanguages();
 
@@ -56,6 +60,7 @@ export class Account {
       unreadCountEnabled: accountConfig.gmail.unreadBadge,
       unifiedInboxEnabled: accountConfig.gmail.unifiedInbox,
       delegatedAccountId: accountConfig.gmail.delegatedAccountId,
+      extensionsLoaded,
     });
 
     this.tabs = new Tabs(accountConfig.id, this.gmail);
--- a/packages/app/extensions.ts
+++ b/packages/app/extensions.ts
@@ -1,6 +1,10 @@
 import path from "node:path";
 import { is } from "@electron-toolkit/utils";
-import { Extensions, findExtensionDirs } from "@meru/electron-extensions";
+import {
+  Extensions,
+  findExtensionDirs,
+  registerNativeMessagingScheme,
+} from "@meru/electron-extensions";
 import { app } from "electron";
 import { serializeError } from "serialize-error";
 import { log } from "@/lib/log";
@@ -26,6 +30,10 @@ function getExtensionDirs() {
 
   return findExtensionDirs(path.join(app.getAppPath(), "extensions"));
 }
+
+// Extension contexts reach the native messaging bridge over a custom scheme,
+// and Electron only takes scheme privileges while modules are still loading
+registerNativeMessagingScheme();
 
 export const extensions = new Extensions({
   extensionDirs: getExtensionDirs(),
--- a/packages/app/extensions.ts
+++ b/packages/app/extensions.ts
@@ -31,6 +31,23 @@ function getExtensionDirs() {
   return findExtensionDirs(path.join(app.getAppPath(), "extensions"));
 }
 
+/**
+ * `MERU_EXTENSIONS_STRIP=content_scripts,declarative_net_request` derives every
+ * extension without those manifest keys, so a run can tell which part of an
+ * extension a page is reacting to. Development only, like the extensions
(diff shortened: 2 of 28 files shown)


timche:
Three fix commits from real-machine testing (macOS, warm profile):

- **403 root cause**: Chromium stores an extension SW's script on first registration and never re-fetches it, so from launch 2 the SW ran a stale facade with a previous launch's bridge token — any account count. (Not the multi-account collision it looked like; both sessions share one token and both register it.) Fix: the loader drops the extension's SW registration before `loadExtension` (`clearStorageData({origin, storages:["serviceworkers"]})` — `clearData` skips `chrome-extension://` origins), using the id derived from `manifest.key`. Origin-scoped: Gmail's own SW registrations are untouched (verified). Side effect worth knowing: before this, **no facade change ever reached an existing profile's SW**.
- Derived copies drop `webRequest`/`webRequestAuthProvider` from permissions — extension-side webRequest is inert in Electron anyway, and this removes the renderer's `NOTREACHED: Module resource "webRequest" not found` bursts entirely (20 → 0) while 1Password still boots, injects content scripts, and reaches the native host.
- Gmail's `loadURL` rejections are now caught and logged (`Failed to load URL`) instead of surfacing as UnhandledPromiseRejection.

The Gmail `ERR_FAILED` itself never reproduced headlessly in any configuration (extension present/absent, 1–2 accounts, cold/warm, both load orders) — Gmail loaded every time. The NOTREACHED burst turned out to fire on successful loads too, from the extension SW process, so its removal is a cleanup of a known-bad path, not a proven fix for the blank window. If it still fails on a real machine, the new `Failed to load URL` log line is the next datum.

timche:
Round 2 on the blank Gmail: not reproducible headlessly even with a faithful analogue of the failing setup — a service-worker-controlled navigation (cached and served via respondWith, over https with 1Password's full content-script set injected, verified `credentialsGet: overridden`), with the extension load racing mid-navigation, one and two accounts. Zero failures across all variants. DNR independently exonerated (its ruleset only touches DoH XHRs).

Two commits added instead:
- Gmail-view failure detail: `render-process-gone` (reason/exitCode) and `did-fail-load`/`did-fail-provisional-load` (errorCode/description/isMainFrame) now logged, so the next failing run says whether the renderer died or the navigation aborted, and with what.
- `MERU_EXTENSIONS_STRIP=<manifest keys>` (dev-only): derives the copy without the named top-level keys, part of the derive stamp so flips re-derive. Verified: `content_scripts` stripped → no page integration, extension still boots.

timche:
Round 3: root-caused. The failing run's log showed the load aborting via Electron's `did-stop-loading` fallback — no `did-fail-load`, no `did-fail-provisional-load`, no `render-process-gone` — i.e. a silent pre-commit abort, 95ms after `Loaded extension` (round 2's failure: 84ms after). In the successful run, Gmail's preload had logged to console *before* `Loaded extension`. So the un-awaited `extensions.setupSession` raced the Gmail navigation: when `loadExtension` completes while the navigation is still provisional, Chromium aborts it; when the navigation commits first, all is well.

Fixed in 2c61884 (Gmail's first navigation waits for the session's extension loads; extension failures log and never block Gmail) and f8e11fb (the initial load retries once if it still never commits, logging `Retrying Gmail load`). The `DidStartWorkerFail …: 5` line seen at startup is benign — it's the dropped SW registration from the freeze fix; Chromium re-registers and boots the SW normally.

PR #852, opened 2026-08-17
let users add sites to an extension's content script clamp

timche:
- New `extensions.additionalSites` config key (extension id → hostnames, default `{}`).
- Shared `normalizeExtensionSiteHostname` (strict: bare lowercase hostname, no ports/paths/wildcards; unit-tested) and `hostnameToMatchPattern` (`https://<host>/*`).
- Main appends the user's sites to the catalog clamp; an unclamped extension never becomes clamped by additional sites.
- Settings: an "Additional Sites" list under each installed, licensed, clamped extension — add via input (Enter submits), remove per row, restart-required toast on change. Covers sign-in through a third-party SSO provider.

Not verified: the footer's rendering in the running app (no display in the sandbox). IDN hostnames must be entered as punycode.

Test plan:
1. Install 1Password → "Additional Sites" appears under it; adding `sso.example.com` lists it and shows the restart toast; a URL, port, or wildcard input shows an error toast instead.
2. After restart, a login page on the added host gets 1Password's inline UI; accounts.google.com still works.
3. Remove the site and restart — the host no longer gets content scripts.

### DIFF
--- a/packages/app/config.ts
+++ b/packages/app/config.ts
@@ -127,6 +127,7 @@ export const config = new Store<Config>({
     "verticalTabs.hideUnreadBadgeWhenActive": false,
     "verticalTabs.showAppLinksBadge": true,
     "extensions.installed": [],
+    "extensions.additionalSites": {},
   },
   migrations: {
     ">=3.4.0": (store) => {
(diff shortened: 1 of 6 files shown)


timche:
Parked for now: whether users should be able to add domains to an extension's clamp at all still needs a decision. The branch stays for when it's picked up.

timche:
Also parked on a second question: whether Meru users sign in through third-party SSO at all — worth evaluating before deciding on user-added domains.

PR #884, opened 2026-08-19
upgrade oxlint and oxfmt

timche:
First of a stack that moves Meru onto `@timche/oxc-configs` 2.0.0, which grows the shared config from 4 rules to around 150 and turns on type-aware linting.

On its own this PR changes nothing. Against the config as it stands today, oxlint 1.79 reports no new errors and oxfmt 0.64 reports no formatting differences — verified before splitting the rest of the work up.

Every PR above this one fixes one group of what 2.0.0 flags. The last one swaps the config in.

## Test plan

1. `bun run lint` — passes with the current four-rule config.
2. `bun run fmt:check` — reports every file already correctly formatted, so no reformatting rides along with the version bump.


timche:

PR #885, opened 2026-08-19
make the type-only and duplicate imports consistent

timche:
`typescript/consistent-type-imports`, `import/no-duplicates` and `import/no-named-as-default` from the new shared config.

- `BrowserWindow` and `WebContentsView` are only used as types, so they take the inline `type` specifier the neighbouring Electron imports already use. Oxlint's own fix hoists them into a separate `import type` statement instead, which isn't how the rest of these files read, so they're written by hand.
- `web-navigation.ts` typed its lazy `require("electron")` with an inline `import()` annotation. It now uses a namespace type import, which is erased the same way and keeps the require lazy — the reason the module doesn't import Electron for value in the first place.
- `accounts.tsx` imported `@meru/shared/schemas` twice, and two files imported zod's `z` as a default import. `import { z }` is what the other two zod importers in the repo use.

## Test plan

1. `bun run types` — passes, confirming the namespace type import in `web-navigation.ts` types `webContents.fromId` the same way `typeof import("electron")` did.
2. `bun test packages/electron-extensions` — passes, so the require stayed lazy and the tests still run outside Electron.

### DIFF
--- a/packages/app/gmail/index.ts
+++ b/packages/app/gmail/index.ts
@@ -17,13 +17,13 @@ import { clamp, wait } from "@meru/shared/utils";
 import type { SupportedWorkspaceApp } from "@meru/shared/workspace-apps";
 import {
   app,
-  BrowserWindow,
+  type BrowserWindow,
   clipboard,
   type Session,
   type WebContentsView,
   type WebContentsViewConstructorOptions,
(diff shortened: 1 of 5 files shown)


timche:

PR #886, opened 2026-08-19
modernize the string, array and undefined checks

timche:
The mechanical half of the new shared config's `unicorn` rules, applied with `oxlint --fix` and reviewed hunk by hunk.

- `String#replace` with a global regex becomes `replaceAll` in ten places. In `generateGmailLabelColorsCss` the two single-character regexes become plain string arguments, which `replaceAll` treats literally.
- `typeof x === "undefined"` becomes `x === undefined` throughout the config migrations.
- `.map(…).flat()` in `useUnifiedInbox` becomes `.flatMap(…)`, two adjacent `patterns.push` calls in the blocker become one, `storages?.join()` spells its `","` out, and `dotIndex < 0` becomes `dotIndex === -1`.
- `loadImage` assigns `onload`/`onerror` handlers, which `unicorn/prefer-add-event-listener` flags and has no fix for; those are `addEventListener` calls now.

No behavior changes — every rewrite here is an equivalent spelling.

## Test plan

1. Open Settings → Gmail → Label Colors and add a label whose name contains a `"` or a `\` — the generated CSS still escapes it, covering the `replaceAll` calls that swapped a regex for a string pattern.
2. Trigger a Gmail notification with a verification code in it — `extractVerificationCode` strips the spaces and dashes through five `replaceAll` calls now.
3. Open a Gmail message with images while the dark theme extension is on — `loadImage` resolves through the new `addEventListener` handlers.

### DIFF
--- a/packages/app/blocker/hosts.ts
+++ b/packages/app/blocker/hosts.ts
@@ -515,7 +515,7 @@ const EMAIL_TRACKER_PATTERNS = [
 ];
 
 function escapeRegExp(value: string) {
-  return value.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
+  return value.replaceAll(/[.*+?^${}()|[\]\\]/g, "\\$&");
 }
 
 function createHostBoundaryPattern(hosts: string[]) {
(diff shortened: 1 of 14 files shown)


timche:

PR #887, opened 2026-08-19
say what each ts-expect-error in the config migrations suppresses

timche:
The new shared config sets `typescript/ban-ts-comment` to `allow-with-description`, and all 20 violations sit in the config migrations.

Each migration already described its first suppression and left the rest bare, so the bare ones now repeat the description of the key they touch. Nothing else changes — every suppression stays exactly where it was.

## Test plan

1. `bun run types` — passes, so no `@ts-expect-error` was left describing a line that no longer errors.
2. Launch with a config from an older version (or delete a migrated key from `config.json` and reopen) — the migrations still run and rewrite the key.


timche:

PR #889, opened 2026-08-19
share the tsconfig bases so toSorted and toReversed type-check

timche:
`unicorn/no-array-sort` and `no-array-reverse` want `toSorted` and `toReversed`, which are ES2023. Every `tsconfig.json` extended `@total-typescript/tsconfig` directly at `lib: ["es2022"]`, so the rewrite didn't type-check.

Rather than repeat a `lib` override in nine files, this adds a `@meru/tsconfig` workspace package with a `node` and a `dom` base. Each extends the `@total-typescript` preset it was already using and raises `lib` to `es2023`; the ten `tsconfig.json` files now extend those. The `tsconfig.test.json` files inherit it through their package's config.

Electron 43 and Bun both ship these methods, so nothing changes at runtime.

Two of the twelve rewrites stop mutating an array in place:

- `Gmail#fetchInboxFeed` reversed `newMailIndexes` before iterating it. Nothing reads the array afterwards, so the order it's left in doesn't matter today.
- `Accounts#createViews` sorted the result of `getAccounts()`, which builds a fresh array on every call, so nothing observed that either.

The other ten already sorted a freshly built array.

## Test plan

1. `bun run dev` — the dev server boots and the app opens, confirming the tsconfig indirection doesn't break the build.
2. Open Settings → Languages — the list is still sorted by label, through `toSorted` now.
3. Receive two or more new mails at once — the notifications still arrive oldest first, which is what reversing the indexes was for.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -101,6 +101,10 @@
       "name": "@meru/shared",
       "version": "0.0.0",
     },
+    "packages/tsconfig": {
+      "name": "@meru/tsconfig",
+      "version": "0.0.0",
+    },
     "packages/ui": {
       "name": "@meru/ui",
       "version": "0.0.0",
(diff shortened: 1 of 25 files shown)


timche:

PR #890, opened 2026-08-19
mark the deliberately unawaited promises with void

timche:
`typescript/no-floating-promises` reports 84 calls whose promise nobody holds on to — menu clicks that open a URL or a file, `init()` at the bottom of the main entry, `queryClient.invalidateQueries` in mutation callbacks, `form.handleSubmit()` in submit handlers.

Every one of them takes a `void` operator. That says the promise is deliberately dropped and leaves behavior exactly as it was: awaiting instead would need the surrounding Electron `click` and React handlers to become async, which they're typed void-returning for, and would only trade this rule for `strict-void-return`.

Rejections stay unhandled, the same as before this PR. Turning them into logged failures is a change worth making on its own terms, not inside a lint sweep.

Applied with `oxlint --fix` and reviewed hunk by hunk; the diff is 84 identical insertions.

## Test plan

1. Help → Website, Help → Source Code, and Advanced → View Logs — each still opens, covering the `openExternalUrl` and `shell.openPath` call sites.
2. Advanced → Clear Cache — the restart dialog still appears afterwards; that call sits inside an already-async click handler and was left unawaited rather than awaited.
3. Add an account in Settings → Accounts — the form still submits, covering `void form.handleSubmit()`.

### DIFF
--- a/packages/app/accounts.ts
+++ b/packages/app/accounts.ts
@@ -48,7 +48,7 @@ class Accounts {
 
     config.onDidChange("gmail.labelColors", () => {
       for (const account of accounts.instances.values()) {
-        account.gmail.applyLabelColors();
+        void account.gmail.applyLabelColors();
       }
     });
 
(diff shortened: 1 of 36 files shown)


timche:

PR #891, opened 2026-08-19
rename the variables that shadowed an outer one

timche:
`no-shadow` reports 22 bindings. Four are in `packages/shared/ms.ts`, a vendored copy of vercel/ms, and are left alone — the last PR in the stack exempts that file. The other 18 are renamed to say what they hold:

- Three `SupportedWorkspaceApp` parameters named `app` shadowed Electron's imported `app`; they're `workspaceApp` now. In the `workspaceApps.openApp` IPC handler that name was already taken by the tab it opens, which is `openedTab`.
- `AccountConfig` values named `account` or `config` shadowed the `accounts` manager and the `config` store; they're `accountConfig` / `accountConfigs`.
- In `fetchInboxFeed`, the feed's response text and a notification's body were both `body`. The response is `inboxFeedXml`.
- The rest are callback parameters shadowing the value they're derived from: `focusedWindow`, `savedSearch`, `text`, `checked`, `value`/`label`, `version`, `res`.

Renames only.

## Test plan

1. Open a workspace app from the launcher in a tab, in a background tab, and in a new window — that IPC handler took the largest rename.
2. Remove an account in Settings → Accounts — the remaining accounts survive and one of them ends up selected, covering the `removeAccount` filter and `every` callbacks.
3. Open the tray menu with unread mail on more than one account — each account still shows its own label and count, and clicking one selects it.

### DIFF
--- a/packages/app/accounts.ts
+++ b/packages/app/accounts.ts
@@ -367,9 +367,9 @@ class Accounts {
 
     const updatedAccounts = config
       .get("accounts")
-      .filter((account) => account.id !== selectedAccountId);
+      .filter((accountConfig) => accountConfig.id !== selectedAccountId);
 
-    if (updatedAccounts.every((account) => account.selected === false)) {
+    if (updatedAccounts.every((accountConfig) => accountConfig.selected === false)) {
       if (!updatedAccounts[0]) {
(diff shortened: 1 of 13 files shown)


timche:

PR #893, opened 2026-08-19
drop the conversions and comparisons that change nothing

timche:
`no-unnecessary-type-conversion`, `no-unnecessary-boolean-literal-compare` and `no-unnecessary-template-expression` — three rewrites that read the same and one `Boolean()` triple that was already boolean.

`no-unnecessary-type-assertion` is left alone here. All 14 of its reports are in test files, where the type information oxlint sees is missing `noUncheckedIndexedAccess` and the assertions it calls redundant are ones `tsc` needs — applying its fix broke `bun run types` in eight places. That's handled in the shared config rather than by deleting the assertions.

## Test plan

1. `bun run types` — passes, confirming `dormantTab.loadOnLaunch` and its two neighbours were already `boolean` and didn't need the `Boolean()` wrapper.
2. Reopen the app with a dormant tab restored from a previous run — it comes back pinned or unpinned as it was, covering the `DormantTab` constructor.

### DIFF
--- a/packages/app/accounts.ts
+++ b/packages/app/accounts.ts
@@ -369,7 +369,7 @@ class Accounts {
       .get("accounts")
       .filter((accountConfig) => accountConfig.id !== selectedAccountId);
 
-    if (updatedAccounts.every((accountConfig) => accountConfig.selected === false)) {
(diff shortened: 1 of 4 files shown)


timche:

PR #894, opened 2026-08-19
remove the conditions that can never change the outcome

timche:
`typescript/no-unnecessary-condition` on the conditions that can only answer one way:

- `!accounts` in the titlebar and `props.onOpenChange &&` in the license dialog were guarding against a value the type says is always there.
- `root.ownerDocument.head?.` and `uniqueErrors?.length` optional-chained through something non-nullish. The `Field` line also had `==` where the rest of the file uses `===`.
- Two `while (true)` loops become `for (;;)`, which is the spelling the rule accepts for a deliberate one.

Two conditions stay, each with a comment saying why and a rule directive naming it:

- `Gmail#loadURL` re-checks `_view` after awaiting the load. `destroy` clears it, which the await gives it time to do, so the narrowing from the check before it no longer holds — the type system just can't see that.
- The extensions page filters curated extensions by `category === "passwordManager"`. `CuratedExtensionCategory` has one member so far, so the comparison is constant until a second one arrives; the filter should outlive that.

The eleven reports in `packages/app/config.ts` are left for the last PR in the stack, which turns the rule off for that file — the migrations read and write keys `Config` no longer has, which is what every `@ts-expect-error` in there is for.

## Test plan

1. Open Settings while the app is still loading its config — the titlebar renders once the config arrives rather than blanking, covering the dropped `!accounts` guard.
2. Activate a license key from Settings → License — the dialog closes on success.
3. Open a Gmail message with a background image while the dark theme is on — `substituteVarFallbacks` and `isGradientLayer` still walk their loops to the end.

### DIFF
--- a/packages/app/gmail/index.ts
+++ b/packages/app/gmail/index.ts
@@ -465,6 +465,9 @@ export class Gmail {
 
     const loaded = await loadUrl(this.view.webContents, this.url);
 
+    // `destroy` clears `_view`, which the load above gives it time to do — the
+    // narrowing the check before it left behind no longer holds.
+    // oxlint-disable-next-line typescript/no-unnecessary-condition
     if (loaded || !this._view || this._view.webContents.isDestroyed()) {
       return;
     }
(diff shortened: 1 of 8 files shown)


timche:

PR #895, opened 2026-08-19
use nullish coalescing where the fallback is about absence

timche:
`typescript/prefer-nullish-coalescing` and `prefer-optional-chain`, on the six expressions where the fallback really is about absence.

Two `if (x === undefined) { x = … }` blocks become `??=`, `cached && cached.fingerprint === …` becomes `cached?.fingerprint === …`, and three `|| ""` become `?? ""` where an empty string and a missing value mean the same thing to the caller.

`buildDarkVariableOverrides` keeps its `||` on the resolved value — an empty custom property has to fall through to the fallback — and drops the trailing `|| ""` instead, since the `if (!value)` right below it already covers the missing case.

The rule reports five more `||` in the renderer that are plain boolean disjunctions, where `false ?? other` would keep the `false` and a control would stop disabling itself. Those are handled in the shared config with `ignorePrimitives: { boolean: true }` rather than here.

## Test plan

1. Open Settings → Extensions and press Check for Updates twice in a row — the second press joins the first check instead of starting another, covering `runningCheck ??=`.
2. Open a Gmail message with a dark theme applied — custom properties that resolve to nothing still fall back to the stylesheet's declared value.

### DIFF
--- a/packages/app/config.ts
+++ b/packages/app/config.ts
@@ -380,9 +380,7 @@ export const config = new Store<Config>({
 
       if (Array.isArray(accounts)) {
         for (const account of accounts) {
-          if (account.workspaceApps === undefined) {
-            account.workspaceApps = { savedTabs: [], bookmarks: [] };
-          }
+          account.workspaceApps ??= { savedTabs: [], bookmarks: [] };
         }
 
         store.set("accounts", accounts);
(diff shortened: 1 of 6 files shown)


timche:

PR #896, opened 2026-08-19
keep the handlers in void-returning slots synchronous

timche:
`strict-void-return` and `no-misused-promises` report async callbacks handed to slots typed as returning void — `ipc.renderer.on` handlers, React `onClick`, Electron menu `click`. A `void` operator doesn't answer these the way it did for the floating promises: the function itself has to stop being async.

Each one becomes a named async function beside its handler, called with `void`:

```ts
async function openAddAccountDialog() { … }

ipc.renderer.on("accounts.openAddAccountDialog", () => {
  void openAddAccountDialog();
});
```

Behavior is unchanged — the work still runs unawaited — and the extracted names say what each handler does.

Two more are their own thing:

- `wait` and a test helper passed `setTimeout(resolve, ms)` as a promise executor's body, which returns the timer id. Both take a block now, which also answers `no-promise-executor-return`.
- `promisify(execFile)` in two places: Node types `execFile` as returning the ChildProcess where `promisify`'s signature wants void. Those keep a directive with the reason.

## Test plan

1. With no license key and the trial expired, use Add Account — the Meru Pro toast still appears instead of the dialog.
2. Settings → Downloads → Change… — picking a folder still raises the restart toast, and cancelling raises nothing.
3. Archive or delete a message from a Meru notification — `gmail.handleMessage` still refreshes the inbox afterwards.
4. Advanced → Reset App… — Cancel leaves everything alone, Reset relaunches.

### DIFF
--- a/packages/app/lib/linux.ts
+++ b/packages/app/lib/linux.ts
@@ -5,6 +5,9 @@ import { join } from "node:path";
 import { promisify } from "node:util";
 import { platform } from "@electron-toolkit/utils";
 
+// Node types `execFile` as returning the ChildProcess, which `promisify`'s
+// signature expects to be void.
+// oxlint-disable-next-line typescript/strict-void-return
 const execFile = promisify(childProcess.execFile);
 
 let linuxWindowControlsEnabled: boolean | null = null;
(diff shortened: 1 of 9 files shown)


timche:

PR #897, opened 2026-08-19
give the lists stable keys and the defaults stable references

timche:
The `react` plugin's rules, which the new shared config turns on.

- **`no-object-type-as-default-prop`** (4): object and function literals as default props are rebuilt on every render, so `AccountForm`, `SavedSearchForm`, `LabelColorForm` and `PopupWindow` now default to a module-level constant.
- **`no-array-index-key`** (3): the sidebar's buttons key on their `path`, and `FieldError` keys on the message it renders — the list is deduplicated by message, so those are unique. The sidebar's separators keep the index, since a separator carries nothing else, with a directive saying so.
- **`button-has-type`** (1): the emoji picker's button gets `type="button"` ahead of its prop spread, so a caller can still override it.

Two React Compiler rules report a library rather than a mistake, and each keeps a directive with the reason: `useReactTable` hands back functions the compiler can't memoize, and wouter's `<Router hook={useHashLocation}>` takes a hook as a value by design.

`find-in-page` had a `biome-ignore` for an effect that deliberately depends on `isActive` alone — typing goes through `debouncedOnFind`, so adding `text` and `onFind` would re-run the search on every keystroke. Biome is gone, so that becomes an oxlint directive; the sidebar had two stale ones too.

## Test plan

1. Open Settings → Accounts → Add Account, close it, and open it again — the form is empty both times, covering the hoisted default.
2. Press Cmd/Ctrl+F, type a query, then close and reopen Find — the search re-runs against what's still in the box.
3. Open the emoji picker from an account label and pick an emoji — it appends rather than submitting the form, which is what the button's type decides.

### DIFF
--- a/packages/renderer/components/app-sidebar.tsx
+++ b/packages/renderer/components/app-sidebar.tsx
@@ -145,14 +145,15 @@ export function AppSidebar() {
             .filter((item) => !item.hidden)
             .map(({ type, label, path }, index) => {
               if (type === "separator") {
-                // biome-ignore lint/suspicious/noArrayIndexKey: Key is acceptable here
+                // A separator carries nothing of its own to key on, and the list
+                // is a fixed one that never reorders.
+                // oxlint-disable-next-line react/no-array-index-key
                 return <Separator key={index} />;
               }
 
               return (
                 <Button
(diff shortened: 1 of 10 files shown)


timche:

PR #898, opened 2026-08-19
throw the right error type and keep the cause

timche:
Four rules about thrown and caught values.

- **`unicorn/prefer-type-error`** (4): a throw that follows a `typeof` check is a `TypeError`. `ConfigSwitchField` and `ConfigSelectField` guarding their config key, the CRX manifest without a version, and the host manifest without `allowed_origins`.
- **`eslint/preserve-caught-error`** (1): the native-messaging framing decoder rethrew a JSON parse failure with the message interpolated but the error itself dropped. It now passes `{ cause: error }` as well, so the stack survives.
- **`typescript/use-unknown-in-catch-callback-variable`** (3): `.catch((error) => …)` and a `then` rejection handler take `unknown`. The one that read `error.message` narrows with `instanceof Error` first, matching the five places in the repo that already do.
- **`eslint/no-new`** (2): `createComposeWindow` returns the `WorkspaceApp` it builds rather than dropping it. The PDF-viewer window can't — `handleWindowOpen` has to answer with a window-open action — so that one keeps a directive saying the instance registers itself.

## Test plan

1. Compose a message from the Meru menu (Gmail → Compose) — the compose window still opens, covering the changed return.
2. Open a PDF from Drive or a Gmail attachment — the viewer still opens in its own window.
3. With 1Password installed, sign in to a Google account through it — the native messaging host still connects, covering the framing decoder's rethrow path.

### DIFF
--- a/packages/app/gmail/index.ts
+++ b/packages/app/gmail/index.ts
@@ -993,7 +993,7 @@ export class Gmail {
   }
 
   createComposeWindow(url: string) {
-    new WorkspaceApp({
+    return new WorkspaceApp({
       accountId: this.accountId,
       url: `${GMAIL_URL}/?extsrc=mailto&url=${encodeURIComponent(url)}`,
       window: { width: 800, height: 600 },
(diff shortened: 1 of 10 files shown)


timche:

PR #899, opened 2026-08-19
spell out the string conversions and drop the deprecated build option

timche:
What's left of the smaller rules.

- **`typescript/restrict-template-expressions`** (5): `${error.cause}` is `unknown` and `${searchParams}` is a `URLSearchParams`. Both say `String(...)` / `.toString()` now, which is what they were relying on implicitly.
- **`typescript/no-deprecated`** (2): `build.rollupOptions` is `build.rolldownOptions` under Vite 8. `enableDeprecatedPaste` stays — Docs cuts and pastes through `document.execCommand`, which Electron gates behind exactly that preference — and keeps a directive with the reason the comment above it already gives.
- **`typescript/await-thenable`** and **`no-confusing-void-expression`** (2): `startElectron()` returns nothing, so the `await` in the dev script came off.
- **`eslint/object-shorthand`** (2) and **`prefer-const`** (1): as reported.
- **`typescript/switch-exhaustiveness-check`** (2): both switches already answer through a `default`, and each now says why, with a directive naming the rule.

  `Account`'s permission handler switches on Electron's permission list, which is long, grows between releases, and is answered default-deny — naming every permission it refuses says nothing.

  `DoNotDisturb#enable` is the more interesting one. Its `default` branch calls `ms(duration)`, and that call is what holds the branch to the fixed durations: a new option `ms` can't parse fails to compile there rather than reaching the branch at runtime. Checked by adding an `"until monday"` option — `tsc` reports `TS2769: No overload matches this call` at that line, so the invariant is enforced rather than assumed.

`oxc/no-map-spread` (6) isn't fixed here. Spreading in a `map` is how the config-state updates are written, and the in-place mutation the rule asks for is the thing they're avoiding; that one is turned off in the shared config.

## Test plan

1. `bun run build:js` — the renderer builds all its entry pages, so `rolldownOptions.input` is being read (verified).
2. `bun run dev` — the dev server comes up and Electron starts, covering the dropped `await` in the watch loop.
3. Turn on Do Not Disturb for 30 minutes, then Indefinitely, then Until Tomorrow — each sets its own expiry and the timer clears between them.
4. Join a Google Meet call and grant camera and microphone — the permission handler still allows `media` and refuses what isn't listed.

### DIFF
--- a/packages/app/account.ts
+++ b/packages/app/account.ts
@@ -124,6 +124,9 @@ export class Account {
         return;
       }
 
+      // Electron's permission list is long, grows between releases, and is
+      // answered default-deny — naming every one it refuses says nothing.
+      // oxlint-disable-next-line typescript/switch-exhaustiveness-check
       switch (permission) {
         case "clipboard-read":
         case "clipboard-sanitized-write":
--- a/packages/app/accounts.ts
+++ b/packages/app/accounts.ts
@@ -22,7 +22,7 @@ class Accounts {
   instances: Map<string, Account> = new Map();
 
   init() {
-    let accountConfigs = config.get("accounts");
+    const accountConfigs = config.get("accounts");
 
     if (!licenseKey.isValid && accountConfigs.length > 1 && accountConfigs[0]?.selected === false) {
       for (const [index, accountConfig] of accountConfigs.entries()) {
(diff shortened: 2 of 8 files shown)


timche:

PR #900, opened 2026-08-19
type the values that were reaching the code as any

timche:
The `typescript/no-unsafe-*` family — values reaching typed code as `any` or as an unresolved type. Most of the 23 reports came from three missing declarations rather than from the code at the call sites.

- **`@types/semver`.** `electron-updater` types `currentVersion` as `SemVer` from `semver`, which ships no types of its own, so `autoUpdater.currentVersion.prerelease` was an unsafe member access on an error type. Adding the types resolves it properly.
- **`declare module "*.css"`.** `packages/preload-gmail` and `packages/preload-workspace-app` declared the module with no body, which makes every CSS import `any` — four reports across the dark theme, the toaster and the Mail preload. They now declare `const content: string` the way `packages/app` already did.
- **The electron-builder hooks.** `beforePack.js` and `afterPack.js` take an untyped `context`; a JSDoc `@param` naming `AfterPackContext` clears eight reports.

The rest are individual:

- Both bridge readers annotate the body as `ReadableStream<Uint8Array>` — the ambient types leave the stream unparameterized, so the chunks arrived as `any`.
- `chrome.contextMenus.create`'s trailing callback narrows through a type predicate, since `typeof x === "function"` only gets you `Function`, which isn't callable safely.
- The slider picks the array case with `typeof value === "object"` rather than `Array.isArray`, which widens a `readonly number[]` to `any[]`.
- The GitHub releases response and the accounts search param are `unknown` and an explicit assertion rather than implicit `any`; zod already validates the first.

The single report left is in `packages/app/config.ts`, which the last PR in the stack exempts along with the rest of the migrations.

## Test plan

1. Switch Settings → General → Update Channel to a pre-release and back — `allowDowngrade` is computed from `currentVersion.prerelease`, which is only properly typed with this change.
2. Open a Gmail message and a compose window with the dark theme on, and open a Google Doc — each loads its injected CSS, covering the three module declarations.
3. Adjust the notification volume slider in Settings → Notifications — it still shows one thumb and sets the value.

### DIFF
--- a/build/hooks/afterPack.js
+++ b/build/hooks/afterPack.js
@@ -1,6 +1,7 @@
 import fs from "node:fs/promises";
 import path from "node:path";
 
+/** @param {import("electron-builder").AfterPackContext} context */
 export default async (context) => {
   if (context.packager.platform.name === "mac") {
     console.log("\n---------\n");
--- a/build/hooks/beforePack.js
+++ b/build/hooks/beforePack.js
@@ -10,6 +10,8 @@ import path from "node:path";
 // provisioning profile embedded in the app authorizes it, and the profile only binds to the app
 // through the application and team identifiers below. `@electron/osx-sign` adds those two itself,
 // but only for sandboxed apps, so they are written out here too.
+// electron-builder hands `beforePack` and `afterPack` the same context.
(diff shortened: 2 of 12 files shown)


timche:

PR #901, opened 2026-08-19
keep the key types through Object.keys and Object.entries

timche:
`Object.keys` and `Object.entries` widen the key back to `string`, because a value can always carry more properties than its type lists. Five places worked around that with an assertion, which `typescript/no-unsafe-type-assertion` reports.

`objectKeys` and `objectEntries` in `@meru/shared/utils` say it once instead, with the reason and a rule directive on the one line that needs it. The records they're used on here are all declared from a literal, so they have exactly the keys they name.

Six assertions across the settings pages and `shared/google.ts` come out, and the `type-fest` `Entries` import goes with them.

## Test plan

1. Settings → Accounts → Add Account — the colour dropdown lists all fifteen colours.
2. Settings → Workspace Apps — the launcher's app picker and the Open in App exclusion list are both fully populated.
3. Open a link to a Google app from Gmail — `workspaceAppsBySubdomain` still resolves it, which is built from `objectKeys` now.

### DIFF
--- a/packages/renderer/routes/settings/accounts.tsx
+++ b/packages/renderer/routes/settings/accounts.tsx
@@ -8,6 +8,7 @@ import {
   type AccountConfigInput,
   accountConfigInputSchema,
 } from "@meru/shared/schemas";
+import { objectEntries } from "@meru/shared/utils";
 import { Badge } from "@meru/ui/components/badge";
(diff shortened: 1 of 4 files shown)


timche:

PR #902, opened 2026-08-19
check the values crossing into typed code instead of asserting them

timche:
The rest of `typescript/no-unsafe-type-assertion`, in the places where untyped data crosses into typed code.

**Checked instead of asserted**, where a check was cheap and honest:

- The `workspaceApp` URL parameter goes through `isSupportedWorkspaceApp` rather than an assertion, so an unrecognised value now reads as none instead of sailing through as a valid app.
- The facade's trailing callbacks narrow through `isExtensionCallback`, and `chrome.runtime` and `runtime.lastError` through `isChromeNamespace` — `typeof x === "function"` only gets you `Function`, and `ChromeNamespace` is `Record<string, unknown>`, so both needed a predicate.
- Native messaging frames off the bridge go through `isNativeMessagingFrame` rather than being asserted wholesale. Anything that isn't a frame closes the port, the same as a disconnect frame — which is what the old assertion did once the value fell through to the `else`.
- `readVarint` and the CSP directive rewrites read their indexed values and handle the missing case, rather than asserting past `noUncheckedIndexedAccess`.
- `InputGroupAddon` narrows its click target with `instanceof Element` — `HTMLElement` would miss a click landing on the `<svg>` inside an addon button, which is what the guard is for.

**Typed at the boundary**, so the call sites don't have to assert:

- `ExtensionBridge.handle` is generic over the body a route expects. The five routes in native messaging and web navigation name their request type instead of casting `body`, and the one assertion lives in `handle` with a comment saying nothing has checked the JSON — the bridge only vouches for who is calling.
- The Toaster's CSS custom properties move to a named constant, since React's `CSSProperties` lists only the known properties.

**Left as assertions**, each with a directive naming the rule and a comment saying why: the six `JSON.parse` results whose fields are checked immediately afterwards, `require("electron")` in the module that loads it lazily, and the two reads of a global that the derived facade writes.

## Test plan

1. With 1Password installed, unlock it in Meru and fill a Google sign-in — that exercises native messaging connect/post/disconnect and web navigation's frame queries, which are the routes that changed shape.
2. Open a workspace app in its own window, then open one from a URL naming an app that doesn't exist — the first loads as before, the second falls back rather than being treated as a real app.
3. Install and remove an extension from Settings → Extensions — the CRX manifest and derive paths still read their versions.

### DIFF
--- a/packages/app/ipc.ts
+++ b/packages/app/ipc.ts
@@ -9,6 +9,7 @@ import { getWorkspaceAppUrl } from "@meru/shared/google";
 import { ms } from "@meru/shared/ms";
 import { GMAIL_TAB_ID } from "@meru/shared/tabs";
 import type { IpcMainEvents, IpcRendererEvent } from "@meru/shared/types";
+import { objectEntries } from "@meru/shared/utils";
 import { workspaceApps } from "@meru/shared/workspace-apps";
 import {
   app,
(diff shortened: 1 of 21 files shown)


timche:

PR #903, opened 2026-08-19
move to oxc-configs 2.0

timche:
The last PR of the stack: `@timche/oxc-configs` 1.0.6 → 2.0, plus `oxlint-tsgolint` for the type-aware rules it turns on. Everything below this one fixed what the new rules report; this switches them on.

Four overrides go in alongside, each for a place where a rule can't be right about the code:

**`jsx-a11y` off.** Meru's renderer is the chrome around Gmail's own interface, not a page of its own.

**`import/no-cycle` off.** 80 cycles across 21 files in `packages/app` — `main`, `accounts`, `ipc`, `menu`, `tabs`, `workspace-app` and the rest reach for each other in both directions. Untangling the singleton graph is its own piece of work, so it's filed in `docs/todo.md` with a pointer here.

**`packages/shared/ms.ts`** is a vendored copy of vercel/ms, kept diffable against upstream. Four rules off for that file alone.

**`packages/app/config.ts`** migrations read and write keys `Config` no longer has, which is what every `@ts-expect-error` in there is for. `no-unnecessary-condition` and `no-unsafe-assignment` judge those reads by the current type, so they can only be wrong about them.

## Not ready to merge

Four of 2.0's rules report code that's correct as written, and their fixes would change what it does — `require-await` and two `no-unnecessary-*` rules on test files, `prefer-nullish-coalescing` on boolean disjunctions, and `no-map-spread` on immutable updates. Those are being handled in the shared config (timche/oxc-configs#3), not here. `switch-exhaustiveness-check` stays at plain `"error"`; its two reports are answered inline in #899.

So this PR still pins `^2.0.0`, and `bun run lint` fails on it until that lands and is released. Once there's a version, the bump is a one-line change and the stack goes green. Verified locally against a build of that branch: lint, types, format and tests all pass.

## Test plan

1. After bumping to the released version: `bun run lint`, `bun run types`, `bun run fmt:check` and `bun test` all pass.
2. `bun run dev` — the dev server boots and Electron starts.
3. Delete the `packages/app/config.ts` override and run `bun run lint` — the eleven migration reports come back, confirming the override is scoped to that file and not hiding anything elsewhere.


timche:

PR #957, opened 2026-08-24
Add a CI job that proves the app boots

timche:
## Summary
Adds `bun run test:boot`, which launches the app, attaches to it over the Chrome DevTools Protocol and checks the renderer actually painted, and a CI job that runs it on every pull request. Nothing caught "Meru doesn't start" before a release until now. The `boot` job passes on `ubuntu-latest` in 54 seconds, 27 of which are a cold container image pull.

## Problem
`ci.yml` runs fmt, lint, types and `bun test`, and none of them launch the app. A change that breaks startup gets as far as a release before anyone notices, and the failure a user sees is the app not opening at all.

The reason there was no such job is that Electron needs a display and a set of GTK, NSS and GBM libraries a runner does not necessarily have. #956 solved that for developers by running the app in the official Playwright image under Xvfb, driven through `bun run dev:headless`. This is the test that groundwork was for.

## Solution
`tests/boot-smoke-test.ts` spawns `bun run dev:headless`, polls for the debugging port, attaches with `playwright-core`, and asserts on the window it finds. The `boot` job in `ci.yml` runs it and uploads a screenshot when it fails.

The test goes through `dev:headless` rather than launching Electron itself, and the CI job runs that same launcher rather than installing Xvfb on the runner. A runner has passwordless sudo, so `sudo apt-get install xvfb` would be lighter, and a job-level `container:` would be more idiomatic for Actions — but both leave CI exercising a launch path nobody uses locally, which lets `scripts/headless` break unnoticed. The image pull was the expected cost of keeping one code path and turned out to be minor: 27 seconds cold, because the runners and `mcr.microsoft.com` are both on Azure. There is deliberately no separate `docker pull` step: naming the image in the workflow as well as in the launcher gives the two somewhere to drift apart, and the pull shows up in the log either way. Recorded in `decisions.md`, which had flagged browsers in CI as a decision of its own.

The assertion is that appearance settings renders, not merely that a window exists. That route needs no Gmail sign-in, and it renders nothing until the config arrives from the main process, so its title appearing proves the renderer bundle loaded, React mounted, hash routing works and an IPC round trip completed. A window that exists but paints nothing still fails. `SettingsTitle` grew a `data-testid` so the check does not match against copy that is free to change.

Two things about the environment are worth knowing, because both are silent traps:

- The dev server treats Electron exiting as the user closing the app and exits `0` itself, so its exit code says nothing about whether the app crashed. The test treats *exiting before the checks finish* as the failure, never the code.
- Bun does not run a dependency's postinstall unless it is trusted, so Electron's binary is absent after any install, with or without `--ignore-scripts`. The job fetches it explicitly and caches the ~124MB download.

Shutdown removes the container rather than signalling the dev server, which spawned the launcher rather than the app and so cannot reliably reach it; with the app gone the dev server exits on its own. Each run gets a fresh `MERU_HEADLESS_HOME`, so a local run starts from the same empty config CI gets. The instance name and CDP port differ from the launcher's defaults, so a smoke test neither disturbs nor is disturbed by a `dev:headless` already running on the machine.

## Try it
No preview deployment applies — this runs a desktop app. The `boot` job on this PR is the change running: https://github.com/zoidsh/meru/actions/runs/32721512979/job/97413741731

On a Linux machine with Docker:

```sh
bun install --frozen-lockfile
(cd node_modules/electron && node install.js)
bun run test:boot
```

It prints `The app booted and the renderer rendered.` and exits `0`. To see the failure path, change the expected title in `tests/boot-smoke-test.ts` to something else: it exits `1`, writes `boot-smoke-test.png`, and lists the debugging targets the app exposed.

## Not verified
- The failure path has only been exercised locally, by breaking the assertion on purpose. The `actions/upload-artifact` step has never actually run, so whether the screenshot arrives as a usable artifact is unconfirmed.
- The Electron cache has only ever missed. This PR's run populated it; nothing has yet restored from it, so the cache key is unproven.
- The 20 minute step timeout was sized for a cold pull before one had been measured. At 54 seconds observed it is now a wide margin rather than a considered number, and it has never been hit.
- The test assumes the runner has network. The app boots to Google's sign-in page in a child `WebContentsView`, and while the assertions only touch the renderer and should hold regardless, a fully offline runner has not been tried.
- An uncaught error in the renderer fails the test. That is our own code rather than Google's, so it should not be noisy, but only the clean case has been observed — no real renderer exception has been through it.
- The screenshot captures the renderer's HTML only. Gmail and workspace apps are separate views the compositor paints above it, so they appear as blank rectangles; that is the protocol, not a broken app.
- Only exercised on Linux, which is all `dev:headless` supports.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,3 +36,27 @@ jobs:
       - uses: oven-sh/setup-bun@v2
       - run: bun install --frozen-lockfile --ignore-scripts
       - run: bun test
+  boot:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+      - uses: oven-sh/setup-bun@v2
+      - run: bun install --frozen-lockfile --ignore-scripts
+      # Bun does not run a dependency's postinstall unless it is trusted, so the
+      # Electron binary is missing after any install, with or without
+      # --ignore-scripts. Fetching it explicitly is the only way to get one.
+      - uses: actions/cache@v4
+        with:
+          path: ~/.cache/electron
+          key: electron-${{ runner.os }}-${{ hashFiles('bun.lock') }}
+      - run: node install.js
+        working-directory: node_modules/electron
+      # Pulls the container image on first use, so allow for that.
+      - run: bun run test:boot
+        timeout-minutes: 20
+      - if: failure()
+        uses: actions/upload-artifact@v7
+        with:
+          name: boot-smoke-test
+          if-no-files-found: ignore
+          path: boot-smoke-test.png
--- a/.gitignore
+++ b/.gitignore
@@ -175,6 +175,7 @@ dist
 .DS_Store
 
 # Meru
+/boot-smoke-test.png
 build/embedded.provisionprofile
 build/entitlements.mac.plist
 build-js
--- a/bun.lock
+++ b/bun.lock
@@ -46,6 +46,7 @@
         "node-machine-id": "^1.1.12",
         "oxfmt": "^0.57.0",
         "oxlint": "^1.72.0",
+        "playwright-core": "^1.62.1",
         "postcss": "^8.5.8",
         "react": "^19.1.1",
         "react-dom": "^19.1.1",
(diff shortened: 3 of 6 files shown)


timche:
Correction to this PR's description and its squash commit, both of which give the wrong reason for the explicit Electron install step.

They say Bun skips untrusted postinstall scripts. That is not what is happening. The `electron` package publishes **no `scripts` field at all** — `node_modules/electron/package.json` has no `postinstall`, so there is nothing for Bun to skip or trust. `--ignore-scripts` was never the cause, `trustedDependencies: ["electron"]` has no effect, and `bun pm default-trusted` already lists `electron` anyway. `bun pm untrusted` reports zero blocked packages.

The binary download is opt-in through the package's own `install-electron` bin. So the step is required regardless of how the install is invoked, and `./node_modules/.bin/install-electron` is the supported way to run it rather than `node install.js` with a `working-directory`.

Two consequences worth recording:

- The job's install step should become `bunx install-electron`, and the `actions/cache` step around it is not earning its place — it missed on both runs so far, and the download takes about two seconds on a runner.
- More importantly, this is not CI-specific. A fresh checkout following the setup in `CLAUDE.md` gets no Electron binary, so `bun run dev` fails until something runs that bin. That is a repo-level gap, not a test one.

`decisions.md` has been corrected and is the accurate record; the commit message on `main` cannot be.

timche:
Correcting the comment above, which got the consequences wrong even though it got the cause right.

Right: the `electron` package publishes no `scripts` field, so bun's trust rules and `--ignore-scripts` are both irrelevant, and `trustedDependencies: ["electron"]` does nothing.

Wrong: I said a fresh checkout cannot `bun run dev`. It can. `electron/index.js` downloads the binary lazily on first `require`, so an ordinary dev run heals itself.

The actual reason the explicit step is needed is narrower and more interesting. `index.js` checks `ELECTRON_OVERRIDE_DIST_PATH` and returns *before* the existence check that triggers the lazy download:

```js
if (process.env.ELECTRON_OVERRIDE_DIST_PATH) {
  return path.join(process.env.ELECTRON_OVERRIDE_DIST_PATH, executablePath || 'electron');
}
if (executablePath) {
  const fullPath = path.join(__dirname, 'dist', executablePath);
  if (!fs.existsSync(fullPath)) {
    downloadElectron();
  }
  ...
```

So `dev:headless`, which sets that variable, is the one path that never self-heals — and the launcher it resolves to then execs `node_modules/electron/dist/electron` inside the container, which has to be there already. Verified both ways: a plain `require('electron')` with no `dist` prints "Downloading Electron binary..." and succeeds; the same require with `ELECTRON_OVERRIDE_DIST_PATH` set resolves to the launcher and leaves `dist` absent.

So the step stays, `bunx install-electron` is the tidier way to spell it, and there is no repo-level setup gap after all. `decisions.md` now carries this version.

PR #968, opened 2026-08-25
Proof only: does the new assertion catch the pre-fix listener ordering

timche:
Throwaway, not for merge. Restores the listener ordering that #966 removes, so CI can say whether the new assertion in `tests/window-bounds.e2e.ts` actually goes red on a runner where Gmail loads over the network. It passes locally either way, because Gmail cannot load in the sandbox and `createView` resolves before a test can look.

Closing as soon as the runners report.


timche:
Answered what it was for: against the old ordering every platform reported zero listeners on the first cold launch, macOS failing all three attempts. The assertion moves to #966. Deleting the branch.

PR #969, opened 2026-08-25
Prototype: can a runner drive the Windows sizing modal loop

timche:
Throwaway, not for merge. Asks whether `PostMessage` of `WM_SYSCOMMAND`/`SC_SIZE` plus posted arrow keys drives the real sizing modal loop on a GitHub-hosted Windows runner — the code path an edge-drag runs inside, which Playwright cannot reach and nothing has ever covered.

Reading the Windows job for whether `will-resize` fires and whether the window actually changed size, then closing.


timche:
Answered what it was for. A hosted Windows runner can drive the real sizing modal loop by posting WM_SYSCOMMAND/SC_SIZE, SetCursorPos works, and with SPI_SETDRAGFULLWINDOWS on the gesture live-resizes — 40 will-resize and 40 resize, the view tracking all of them. Landed as #971.

PR #970, opened 2026-08-25
Mutation: does the native drag test catch a view that only updates on release

timche:
Throwaway, not for merge. The account views are laid out on `resized` instead of `resize`, which is the reported symptom exactly — frozen through the drag, snapping on release.

If `tests/native-resize.e2e.ts` does not go red here, it is not a detector of the bug it is named for. Reading the Windows job and closing.


timche:
Answered what it was for. With the layout moved to resized, the drag test fails on 40 of 40 resizes, samples showing the window at 708, 716, 724, 732 wide against a view stuck at 1024. The test is a detector rather than a passenger, and it landed as #971.

PR #972, opened 2026-08-25
Take the maximized content size from the display

timche:
## Summary
Maximizing on Windows left the account view two pixels short at the bottom, because Chromium reserves two pixels for a possible auto-hide taskbar while it asks the system asynchronously whether there is one, then recalculates the frame without a window resize behind it. The main window's own page fills the content area, so its resize is the signal Electron does not emit; it now tells the main process to lay the views out again. The maximize assertion tightens back to zero.

## Problem
Maximizing on Windows leaves the account view short at the bottom — view height 678 against a content height of 720. Reproduced on `windows-latest`, three attempts of three.

Timing it and then reading Chromium says exactly why. `HWNDMessageHandler::GetAppbarAutohideEdges` asks the system which screen edges hold an auto-hide taskbar **asynchronously**: it returns a stale answer and schedules a callback. While it waits, `WM_NCCALCSIZE` reserves `kAutoHideTaskbarThicknessPx` — two pixels — along each such edge, so a maximized window is not treated as a fullscreen app and the taskbar stays reachable. When the answer arrives, `OnAppbarAutohideEdgesChanged` calls `SetWindowPos` with `SWP_FRAMECHANGED` on the same rectangle and the frame is recalculated two pixels larger.

Measured on the runner:

    +4ms   EVENT resize
    +33ms  EVENT maximize
    +33ms  bounds 1040x736  content 1024x718  view 1024x678
    +43ms  bounds 1040x736  content 1024x720  view 1024x678

The window bounds never move, and Electron gates `resize` on the window bounds (`NativeWindowViews::OnWidgetBoundsChanged`), so nothing is emitted for that last change. Anything laid out on the event is two pixels short and is never told otherwise. A probe sampling `maximize`, `unmaximize` and a `setImmediate` after each read 718 at every point, which ruled out deferring a turn.

## Solution
- The main renderer reports its own resize; the main process lays the views out again on it

The renderer's page fills the content area, which makes its resize the missing signal, and it is right by construction rather than by premise: it fires when and only when the content area actually changes, whatever moved it. It carries no size — the main process re-reads the bounds itself, which keeps one source of truth and stays correct whatever the page's zoom is doing.

Laying out again is idempotent, so the resizes this shares with the window's own event cost a second pass and change nothing. That is the price for not having to tell the two cases apart, and it is why no attempt is made to.

**A rejected first attempt is worth recording.** This originally read the display's work area when maximized, on the premise that a maximized window's content area is that work area. Review found the hole and Chromium's source confirms it: an auto-hide taskbar reserves no work area, so Windows reports the full monitor while the two-pixel inset above is deliberately kept. The substitution would have run the view two pixels past the client area and clipped Gmail — trading a gap everyone sees for content loss a minority do. The timer variant was rejected earlier still: re-reading after a delay goes green on a runner and becomes a latent flake on slower hardware.

## Not verified
- An auto-hide taskbar, and multiple monitors. CI has one screen with a fixed taskbar. The mechanism no longer depends on either — the renderer observes the content area whatever produced it — but neither is exercised.
- The extra IPC message per renderer resize. During a drag that is one per frame, and the layout it triggers duplicates the one the window's own `resize` already did. Idempotent and cheap, but not measured.


timche:
Superseded by #974. This closed when its base branch was deleted with #966's merge, and GitHub will not reopen a PR whose base is gone. Same branch, same commits rebased onto main, same body.

PR #974, opened 2026-08-25
Take the two pixels from the page that has them

timche:
## Summary
Maximizing on Windows left the account view two pixels short at the bottom, because Chromium reserves two pixels for a possible auto-hide taskbar while it asks the system asynchronously whether there is one, then recalculates the frame without a window resize behind it. The main window's own page fills the content area, so its resize is the signal Electron does not emit; it now tells the main process to lay the views out again. The maximize assertion tightens back to zero.

## Problem
Maximizing on Windows leaves the account view short at the bottom — view height 678 against a content height of 720. Reproduced on `windows-latest`, three attempts of three.

Timing it and then reading Chromium says exactly why. `HWNDMessageHandler::GetAppbarAutohideEdges` asks the system which screen edges hold an auto-hide taskbar **asynchronously**: it returns a stale answer and schedules a callback. While it waits, `WM_NCCALCSIZE` reserves `kAutoHideTaskbarThicknessPx` — two pixels — along each such edge, so a maximized window is not treated as a fullscreen app and the taskbar stays reachable. When the answer arrives, `OnAppbarAutohideEdgesChanged` calls `SetWindowPos` with `SWP_FRAMECHANGED` on the same rectangle and the frame is recalculated two pixels larger.

Measured on the runner:

    +4ms   EVENT resize
    +33ms  EVENT maximize
    +33ms  bounds 1040x736  content 1024x718  view 1024x678
    +43ms  bounds 1040x736  content 1024x720  view 1024x678

The window bounds never move, and Electron gates `resize` on the window bounds (`NativeWindowViews::OnWidgetBoundsChanged`), so nothing is emitted for that last change. Anything laid out on the event is two pixels short and is never told otherwise. A probe sampling `maximize`, `unmaximize` and a `setImmediate` after each read 718 at every point, which ruled out deferring a turn.

## Solution
- The main renderer reports its own resize; the main process lays the views out again on it

The renderer's page fills the content area, which makes its resize the missing signal, and it is right by construction rather than by premise: it fires when and only when the content area actually changes, whatever moved it. It carries no size — the main process re-reads the bounds itself, which keeps one source of truth and stays correct whatever the page's zoom is doing.

Laying out again is idempotent, so the resizes this shares with the window's own event cost a second pass and change nothing. That is the price for not having to tell the two cases apart, and it is why no attempt is made to.

**A rejected first attempt is worth recording.** This originally read the display's work area when maximized, on the premise that a maximized window's content area is that work area. Review found the hole and Chromium's source confirms it: an auto-hide taskbar reserves no work area, so Windows reports the full monitor while the two-pixel inset above is deliberately kept. The substitution would have run the view two pixels past the client area and clipped Gmail — trading a gap everyone sees for content loss a minority do. The timer variant was rejected earlier still: re-reading after a delay goes green on a runner and becomes a latent flake on slower hardware.

## Not verified
- An auto-hide taskbar, and multiple monitors. CI has one screen with a fixed taskbar. The mechanism no longer depends on either — the renderer observes the content area whatever produced it — but neither is exercised.
- The extra IPC message per renderer resize. During a drag that is one per frame, and the layout it triggers duplicates the one the window's own `resize` already did. Idempotent and cheap, but not measured.


timche:
Parked. Not breaking, and the resize bug this was found alongside turned out to be a different one — the startup listener gap in #966, which Tim has confirmed fixed on Windows.

This branch holds the second of two attempts and Tim reports the gap persists on Windows with an auto-hide taskbar, which fits that case being a permanent 2px reserve rather than a late correction: Chromium subtracts `kAutoHideTaskbarThicknessPx` per auto-hide edge from a maximized window's client rect on purpose, so the window is not taken for a fullscreen app. A signal that waits for a correction has nothing to hear.

Open question before a third attempt: whether the auto-hide case is Meru's to fix at all, since that strip is non-client area a child of `contentView` cannot reach.

Root cause, both rejected approaches and the reasoning are in `docs/todo.md`.

timche:
Closed for now, not abandoned. The branch `maximized-view-gap` stays, and its base is `main`, so this reopens cleanly when it is picked up.

Where it stands: this fixes a real, permanent 2px gap when the taskbar is not auto-hidden — the content area settles from 718 to 720 after the last `resize`, leaving the view short with nothing to correct it. It is correctly inert under auto-hide, where the 2px reserve is permanent and Meru's layout is already right.

Root cause, both attempted approaches, and the measurements separating the two cases are in `docs/todo.md`.

PR #975, opened 2026-08-25
Probe: a maximize against an auto-hide taskbar

timche:
Throwaway, not for merge. A 2px bottom gap was reported on Windows with an auto-hide taskbar, on a build carrying #974's renderer-signal fix.

That fix catches a late correction to the content area, so it can only work if there is one. Chromium subtracts `kAutoHideTaskbarThicknessPx` from a maximized window's client rect for every auto-hide edge (`hwnd_message_handler.cc:2590`), deliberately, so the window is not taken for a fullscreen app — which would be a permanent reserve rather than a correction, and nothing for the fix to hear.

This turns auto-hide on with `SHAppBarMessage(ABM_SETSTATE, ABS_AUTOHIDE)`, maximizes, and records the timeline. It restores the previous state in a `finally`.

Closing as soon as it reports.


timche:
Answered what it was for. With auto-hide on, the work area reports the full monitor (1024x768) while the maximized content settles at 1024x766 and never changes — a permanent 2px reserve, not a correction — and the account view fills it exactly at 726. So Meru's layout is already right under auto-hide, the strip is non-client area Chromium keeps on purpose, and #974 is correctly inert there. It also confirms by measurement that the rejected work-area approach would have sized the view 728 against a 766 content area and clipped Gmail. Recorded in docs/todo.md.

PR #986, opened 2026-08-26
Reach the one worker from an extension page too

timche:
## Summary
An extension page in a content-script-only session can now reach the one shared service worker, which is what the action popup needs — it is where 1Password keeps its unlock UI, and in any account but the first there was no worker in its session to talk to. The derive writes the runtime proxy's shim into every page of that copy, widens the page's own content security policy so the shim can reach the bridge, and the relay hands the worker the sender Chrome would have given it. Still off by default behind `MERU_EXTENSIONS_SHARED_INSTANCE`, and still never run against 1Password itself.

## Problem
[PR #985](https://github.com/zoidsh/meru/pull/985) put one shared extension instance behind that switch: the first session set up keeps the whole extension and its single service worker, every later session gets a content-script-only copy, and a shim in that copy's content scripts relays their `chrome.runtime` messaging to the one worker. It deliberately left one gap. The action popup is an extension page, not a content script, and it opens in whichever account session the user is looking at — so in a content-script-only session its `chrome.runtime` messaging had nothing to reach. The popup is 1Password's unlock UI, which made this the one thing standing between #985 and a real 1Password test.

It turned out to be wider than the popup. 1Password's inline UI — `inline/menu/menu.html`, `notification`, `modal`, `universal-sign-on` — is extension pages loaded as iframes inside web pages, and those are on the fill path, not just the unlock path.

## Solution
- The shim already knew how to relay and only ran in content scripts. The derive now also writes it into every extension page of a content-script-only copy, behind the facade and ahead of the page's own scripts, which is the injection vehicle extension pages already had for the facade.
- A page's own `<meta>` policy governs it alongside the manifest's, and the stricter one decides, so 1Password's popup — `default-src 'none'`, no `connect-src` — refuses the bridge however wide the manifest was made. Measured on Electron 43: the fetch fails with "Refused to connect because it violates the document's Content Security Policy" before any handler is asked. `derive/html.ts` widens the page policy the way `allowConnectSource` already widens the manifest's, reusing that function, and only for this role — so the ordinary copy is still derived byte for byte as it was.
- A top-level extension page is no tab. Chrome hands an action popup's messages a sender of `id`, `url` and `origin` alone, so the reconstruction stops before the frame lookup rather than reporting the popup's own `WebContentsView` as a tab. An extension page in a subframe is the other case and keeps the host tab and its frame id, which is what the inline frames need. No new field was required: the shim already reports `isTopFrame`, and a top-level document on the extension's own scheme is an extension page.
- `new URL().origin` answers `"null"` for `chrome-extension:` outside a browser process, and 1Password compares `sender.origin` against its own `getURL("")` and refuses anything else, so extension URLs get the origin Chromium would have given them.
- `DERIVE_VERSION` goes 6 → 7, since what the derive writes changed and the source tree's hash cannot see a change that lives in this repository.

Why this way and not the two obvious alternatives. **Opening the popup in the worker's session** would need no proxying at all, but Electron implements `chrome.tabs` natively and per-session, so the popup would enumerate the wrong account's tabs and fill into the wrong page; it also needs a second seam in `packages/app`, and it does nothing for the inline frames, which are extension pages inside web pages and cannot move sessions. **Giving the bridge scheme `bypassCSP`** would fix the policy problem in one line, but it changes default behavior for every session and hands the scheme to every document in it, where today only the token is the gate.

There are no changes in `packages/app`: the whole feature still hangs off the one `sharedInstance` option. Removability was proven the way #985 proved it — deleting that option leaves `bun run types`, `bun run lint` and `bun test` green, and the derive output of a copy without the option is byte-identical to this branch's base, the stamp's `deriveVersion` alone excepted, which is what forces the re-derive.

What this does **not** add is a way for the worker to reach a page it did not hear from first. `runtime.onMessage` in a page stays native, so a worker's `sendMessage` broadcast crosses no session — the same gap #985 has for content scripts and `tabs.sendMessage`. Ports carry both ways, which is what 1Password's messaging library opens.

## Try it
There is no preview deployment for an Electron app, so this is a local run: two accounts, 1Password installed, then `MERU_EXTENSIONS_SHARED_INSTANCE=1 bun run dev`. Select the second account, open the extensions button in the titlebar and pick 1Password. Before this branch its popup sits there with nothing behind it; after, it reaches the worker in the first account's session, so an unlock in either account is an unlock in both.

Without 1Password, the mechanism is visible with any MV3 extension that has a `manifest.key`, an `action.default_popup` and a background worker that answers `runtime.onMessage` — drop it in the gitignored `extensions/` folder. A key matters: without one the two derived copies get different generated ids and the relay never matches them, which is a #985 limitation this branch does not change.

## Not verified
- **The four flows that must keep working — log in with a password, log in with a passkey, create a password, create a passkey — and shared unlock across accounts.** They need a real signed-in Google account and the 1Password desktop app, neither of which exists in the sandbox this was written in. macOS and Windows were also unavailable, so this is Linux-only.
- Whether 1Password's popup needs any worker-initiated `runtime.sendMessage` broadcast, which this branch does not carry. Its messaging library talks over named ports, which do work, but only a real run says whether anything depends on the broadcast path.
- What was verified, in a bare Electron harness against the loader with two sessions and a purpose-built extension whose popup carries 1Password's own strict page policy: a popup in the content-script-only session completed a `sendMessage` round trip and a named-port echo against the worker in the other session, with the sender `{id, url, origin}` and no `tab`; the same extension's page embedded as an iframe in a web page in that session got the host tab and its frame id; and the worker session's own popup still went the native path untouched.

### DIFF
--- a/packages/electron-extensions/derive/html.test.ts
+++ b/packages/electron-extensions/derive/html.test.ts
@@ -1,11 +1,11 @@
 import { describe, expect, test } from "bun:test";
-import { injectFacadeScript } from "./html";
+import { allowPageConnectSource, injectPageScripts } from "./html";
 
-describe("injectFacadeScript", () => {
-  test("runs the facade before the page's own scripts", () => {
-    const page = injectFacadeScript(
+describe("injectPageScripts", () => {
+  test("runs the loader's scripts before the page's own", () => {
+    const page = injectPageScripts(
       '<!doctype html><html><head><title>Popup</title><script src="/popup.js"></script></head></html>',
-      "/chrome-facade.js",
+      ["/chrome-facade.js"],
     );
 
     expect(page).toStartWith(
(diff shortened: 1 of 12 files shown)


timche:
Folded into #985 rather than stacked. The split stopped making sense once the extension-page gap turned out to cover 1Password's inline menu, notification and modal frames, which sit on the fill path — one mechanism cut in half rather than two changes. Its commit is in #985 with both reviews' findings applied on top, and everything this body said about page-level `<meta>` CSP is carried over there.

PR #990, opened 2026-08-26
Fix three release-risk bugs and the release skill's broken precondition

timche:
## Summary
Four small fixes the [release-risk review](https://github.com/zoidsh/meru/compare/v3.59.0...main) flagged as pre-release work: leaving the Experimental update channel no longer throws, a stale tab id no longer wedges the app in a repeating error dialog, a woken tab whose history fails to restore now loads its page instead of showing a blank view, and `/release` no longer stops at a precondition naming a workflow that was deleted. Three of the four are hibernation- or channel-related dead ends a user cannot get out of without relaunching. Unit tests cover the two fixes with a seam; the tab guard has none.

## Problem
Four independent bugs, all landed since v3.59.0 and all worth closing before the release rather than after it.

**Switching Experimental → Stable in one session throws.** `updater.ts` assigned `autoUpdater.channel = null` for stable, and electron-updater's setter refuses anything but a non-empty string once a channel has been assigned — which it has, the moment the user picks Experimental. The `ERR_UPDATER_INVALID_CHANNEL` escapes the `config.onDidChange` listener as an uncaught main-process exception (there is no `uncaughtException` handler), so the user gets an Electron error dialog, and `allowPrerelease`, `allowDowngrade` and the check that follow never run — they stay on Experimental until the next launch. That defeats [Immediate downgrade on leaving the beta channel](https://github.com/zoidsh/meru/pull/843) and the release-channels doc's "the channel applies live".

**A notification click for a hibernated tab wedges the app.** The workspace-app notification closure captures a tab id; hibernating that tab swaps it for a `DormantTab` with a new UUID, so the captured id names nothing. `activateTab` had no not-found guard and fell through to making that dead id the active one — and the `activeTab` getter throws on an id it cannot resolve, from the menu refresh, from the selected view refresh, and from every one-minute hibernation sweep after it. Blank content area plus an error dialog that returns every minute until the user clicks a live tab. The missing guard predates the range; hibernation is what makes ids go stale.

**A failed history restore leaves a blank view.** Waking a hibernated tab restores its navigation history. `restoreNavigationHistory` already answers whether the page arrived, and the caller dropped that answer — a restore that never commits left a blank view under a tab still wearing its old title.

**`/release` stops at its own precondition.** `8ff4181` deleted `.github/workflows/build.yml` when its jobs moved into `ci.yml`, but the skill still listed the file as one that must be present and still required a non-failing `build.yml` run.

## Solution
- **Updater** — assigns `"latest"` for stable instead of `null`. That is electron-updater's own default channel name and resolves the same `latest*.yml` an unset channel does, so stable resolution is unchanged, with one case it repairs: unset, the GitHub provider falls through to the channel baked into the build's `app-update.yml`, which on an Experimental build is `alpha` — so stepping back to stable asked a stable release for `alpha-mac.yml` and got a 404 with no fallback. The alternative of guarding the assignment was rejected because it leaves that wrong-feed case standing and leaves `_channel` holding `alpha` on a user who has left the channel. The three properties move into a pure `resolveUpdateChannel` in `lib/`, following `lib/hibernation.ts`, so the never-null invariant has somewhere to be tested — the updater itself reaches too far into Electron to test directly.
- **Tabs** — `if (!activatedTab) return;` at the top of `activateTab`, matching the shape `moveTab` already uses. The click still brings the window forward and selects the account, with the hibernated tab in the strip to be opened. Waking the dormant successor instead would need a back-reference the swap doesn't keep, and is a scope question rather than a fix (see **Not verified**).
- **Hibernation wake** — the restore-or-load choice moves into `loadUrlOrRestoreNavigationHistory`, next to the two functions it picks between, where it is unit-testable; the caller sits inside `WorkspaceApp`'s view construction and is not. The fallback is deliberately narrower than "the restore failed", because a restore is reported failed when it is *superseded* as well: Chromium aborts the entry being restored the moment another navigation takes over — a redirect, a client-side route, a link clicked into the waking tab — and loading over that would lose a page the user had reached, which is worse than the blank view being fixed. Having committed nothing is what tells the two apart, and a view created for a waking tab has committed nothing by definition, so an empty `getURL()` is the blank case exactly. Two other states are left alone for the same reason: a view destroyed mid-restore, because loading into one throws rather than answering false, and a history of no entries, which is what a tab that hibernated before its first load committed carries — `restorableNavigationHistory` always answers an object, `{ entries: [], index: -1 }` in that case, and `canHibernateTab` lets such a tab hibernate because `WorkspaceApp.url` falls back to the app's own.
- **Release skill** — the build check is kept rather than dropped. `ci.yml`'s `e2e` job builds the app with electron-builder and launches it on macOS, Windows and Linux, so it rides on the `ci.yml` run already required to be green at the exact `HEAD` being released. That is stricter than what it replaces: `build.yml` carried `paths-ignore: "**.md"` and was allowed to lag `HEAD`. What the merged check gives up is written into the skill rather than assumed — `e2e` builds `--dir` and unsigned, so installer packaging and macOS signing still run for the first time in the release itself. `decisions.md` gets a new entry marking the old two-workflow entry superseded. Also fixes the `docs/decisions.md` path in the same skill file, left behind by the same deletion wave.

## Try it
No preview deployment for a desktop app — `bun run dev`, then:

- **Updater**: Settings → Updates → set channel to Experimental, then back to Stable, same session. Before: an Electron error dialog and the select stuck on Experimental. After: it switches, and one update check fires. (`is.dev` short-circuits `checkForUpdates`, so a packaged build is what shows the check itself.)
- **Hibernation**: Settings → Workspace Apps → hibernation on with the shortest timeout, mark a tab, let it idle past the timeout, then click a notification that tab fired. Before: blank content area and an error dialog every minute. After: the window comes forward on the right account and nothing throws.
- **Release skill**: `/release` now gets past its preconditions.

## Not verified
- The superseded-restore and empty-entries cases above are covered by unit tests against a stubbed `WebContents`, not against Chromium. That `navigationHistory.restore` rejects on `ERR_ABORTED` is read off Electron's own contract — it "rejects if the page fails to load (see `did-fail-load`)" — rather than observed; the guard is correct either way, since it only ever declines to load over a view that is already showing something.
- No manual run of any of the three app fixes — the sandbox has no display, so all three are reasoned from the code and covered only by unit tests. The updater path in particular is only exercised as a pure function; the real `autoUpdater` was never driven.
- The tab guard has no test. `Tabs` reaches into Electron throughout and there is no harness for it; the two fixes with a seam got one (13 new unit tests across `lib/update-channel.test.ts` and `lib/load-url.test.ts`).
- Whether a notification click should wake the hibernated tab's dormant successor rather than doing nothing is open — the review asked for the guard alone, and the successor mapping is real scope.
- `bun run types`, `bun run lint`, `bun run fmt:check` and `bun test` (550 pass, 0 fail) were run. The e2e and perf suites were not — they need a display.

### DIFF
--- a/.claude/skills/release/SKILL.md
+++ b/.claude/skills/release/SKILL.md
@@ -13,13 +13,12 @@ The version commit goes straight onto `main` — no branch, no PR. Publishing th
 Check all of these first. If one fails, report it and stop — never work around it.
 
 - On `main`, clean and up to date: `git status --porcelain` empty, then `git checkout main && git pull --ff-only`.
-- The workflow names below are real: `.github/workflows/` holds `ci.yml` and `build.yml`. Check that before running either command — GitHub answers a renamed workflow with its pre-rename runs instead of an error, so a stale name here reads as "no run yet" on every release and never verifies anything.
+- The workflow names below are real: `.github/workflows/` holds `ci.yml` and `release.yml`. Check that before running any `gh run list --workflow=` command — GitHub answers a renamed workflow with its pre-rename runs instead of an error, so a stale name here reads as "no run yet" on every release and never verifies anything.
 - The last `ci.yml` run is for the current `HEAD` and passed: `gh run list --workflow=ci.yml --branch=main -L 1 --json headSha,status,conclusion,url`.
   - `headSha` must equal `git rev-parse HEAD`. A `HEAD` with no run yet, or a run still `in_progress`, means waiting — say which and ask whether to wait for it.
   - Any `conclusion` other than `success` means `main` is broken. Report the run URL and stop.
-- The newest `build.yml` run on `main` is not a failure: `gh run list --workflow=build.yml --branch=main -L 1 --json headSha,status,conclusion,url`. A green `ci.yml` says nothing about whether the app compiles, and `release.yml` builds all three platforms — a broken build fails the release where it costs the most.
-  - A `conclusion` of `failure` means the release build will fail too. Report the run URL and stop. A run still `in_progress` means waiting — ask whether to wait for it.
-  - Its `headSha` may lag `HEAD`, because `build.yml` ignores commits that only touch Markdown. Don't require a match, and don't wait for a run that will never start.
+  - That run is also the build check. `ci.yml`'s `e2e` job builds the app with electron-builder and launches it on macOS, Windows and Linux, so a green run at `HEAD` is what says the app still compiles on all three — the thing `release.yml` does next, at the most expensive place for it to fail. There is no second workflow to query; `build.yml` was deleted when its jobs moved here.
+  - What it still doesn't cover: `e2e` builds `--dir` and unsigned, so installer packaging and macOS signing run for the first time in the release itself. That is a known risk of every release, not something to check here.
 - There is something to release: the range from the last release's tag to `HEAD` is non-empty. For a stable release that's the last stable tag, from `gh release list --exclude-pre-releases -L 1`. For an experimental release it's the last release of any kind, from `gh release list -L 1`.
 
 ## Choose the bump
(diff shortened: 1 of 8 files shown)


timche:
Split into four independent PRs, one per fix — they touch disjoint files and none depends on another, so they target main and merge in any order:

- #992 — release skill build precondition
- #993 — `activateTab` not-found guard
- #994 — stable update channel
- #995 — hibernation wake fallback

Closing this in favor of those. Branch `pre-release-fixes` deleted.

PR #998, opened 2026-09-01
Stop loading 1Password in development on macOS

timche:
## Summary
A development build can't reach the 1Password desktop app, so the extension loaded into every account session and sat there unable to unlock. This drops it from both development sources on macOS, with `MERU_DEV_ENABLED_EXTENSIONS=<id>` to load it anyway for work on the extension layer. macOS alone: Windows is untested and left loading.

## Problem
The 1Password desktop app verifies the calling browser's code signature before it will talk to it, and `bun run dev` runs an unsigned `electron`, so the native-messaging host answers `BrowserVerificationFailed` and `UnknownBrowser`. The extension still loads, so every account session in development carries a 1Password that can never unlock.

It costs more than nothing. While the extension is loaded, `workspace-app.ts:179` holds back the passkey dialog, on the grounds that 1Password's `navigator.credentials` override makes it moot — so development gets a 1Password that can't sign you in and no fallback dialog either.

## Solution
- `getDevDisabledExtensionIds` in `packages/app/extensions.ts` names the curated extensions a development build refuses to load, which on macOS is 1Password. `MERU_DEV_ENABLED_EXTENSIONS=<id>,<id>` loads them anyway, comma-separated so it isn't a 1Password switch.
- Both development sources are gated, the `extensions.installed` opt-in and the repository's `extensions/` folder, because either can carry a copy and gating one leaves the other loading.
- Matching the folder's copy needs its id, so `readExtensionDirId` in `@meru/electron-extensions` reads it from the manifest `key` — the value the download script injects and the same one Chromium derives the id from. A folder with no key has no id before it loads, so it loads; that is a hand-placed extension, which is deliberate by definition.
- macOS alone rather than every platform. Windows runs the same Add Browser trust flow and probably fails the same way, but nobody has run it, so guessing would disable something that may work. Linux trusts a binary name through `/etc/1password/custom_allowed_browsers` rather than a signature, so a development run can be allowlisted there.

The `extensions/` folder stays as it is. It is the only route to an extension outside the curated catalog or one with local edits, which install-from-settings refuses.

## Try it
On a Mac, `bun run dev` with 1Password installed from the Extensions settings page: the account sessions come up without it, and the passkey dialog appears again on a Google sign-in page. `MERU_DEV_ENABLED_EXTENSIONS=aeblfdkhhhdcdjpifhhbdiojplfjncoa bun run dev` brings it back. No preview deployment applies.

## Not verified
- Nothing was run in the app. This sandbox is Linux with no 1Password desktop app, so the gate is verified by unit tests and reading only — `bun test`, `bun run types`, `bun run lint` and `bun run fmt:check` all pass.
- Whether a development build reaches the 1Password host on Windows is still unknown, which is why Windows keeps loading the extension. Recorded as an open question in the feature doc.

### DIFF
--- a/packages/app/extensions.ts
+++ b/packages/app/extensions.ts
@@ -1,5 +1,5 @@
 import path from "node:path";
-import { is } from "@electron-toolkit/utils";
+import { is, platform } from "@electron-toolkit/utils";
 import {
   createSharedExtensionInstance,
   Extensions,
@@ -9,10 +9,15 @@ import {
   type LatestExtensionInstall,
   pruneDerivedExtensions,
   pruneExtensionVersions,
+  readExtensionDirId,
   registerExtensionBridgeScheme,
   uninstallExtension,
 } from "@meru/electron-extensions";
-import { curatedExtensions, isCuratedExtensionId } from "@meru/shared/extensions";
+import {
+  curatedExtensions,
+  isCuratedExtensionId,
+  ONEPASSWORD_EXTENSION_ID,
+} from "@meru/shared/extensions";
 import { ms } from "@meru/shared/ms";
 import type { ExtensionUpdateResult, InstalledExtensionState } from "@meru/shared/types";
 import { app } from "electron";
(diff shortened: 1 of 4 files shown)


timche:
Closing undelivered: a 1Password that can't reach the desktop app in development is harmless in practice, so the gate isn't worth the code. The duplicate-extension-id finding this turned up is real and being fixed separately, carrying `readExtensionDirId` with it.

PR #1053, opened 2026-09-03
Upgrade Electron to 44.3.0

timche:
## Summary

Takes Electron from 43.7.0 to 44.3.0 — Chromium 150 → 152 — and migrates the main process onto 44's promise-returning `clipboard` API. `LSMinimumSystemVersion` declares macOS 13, which is the half of the Monterey cut that main does not already have: the updater gate landed separately in `packages/app/updater/`, and this is what makes a Mac below Ventura refuse the app with an explanation rather than launch it into an unsupported Chromium.

Two things came out of the work rather than the version bump: a verification code could be deleted after a failed clipboard write, and `mock.module` leaks across test files without `--isolate`. Both are their own commits.

What this branch does not settle is Gmail. Its three DOM-coupled features can only be exercised against a real signed-in session, and no Mac or Windows machine was available.

## Node goes backwards by one patch

43.7.0 picked up Node 24.21.0; the 44 line is still on 24.20.0. That is not a security release and names no CVE, but it does carry OpenSSL 3.5.8 and NSS 3.126 root certificates, which this gives back. Those reach the Node-side HTTP the updater and the API client use, not Chromium's own stack, and the 44 line catches up on its own.

## What 44 breaks, against the code

`clipboard` is rebuilt on the W3C API, so `writeText` returns a promise. All seven main-process call sites go through one `copyText` helper that awaits the write, because each is a menu click or a notification handler with no caller to hand a rejection to and Meru registers no `uncaughtException` handler.

The rest was re-checked rather than assumed, and reaches nothing: renderers already use `navigator.clipboard`, there are no `net` module call sites, no `select-client-certificate` handler, only `openAtLogin` is read from login item settings, nothing swaps ANGLE libraries, and the builder targets x64 and arm64 only. `bun run types` passes with no change beyond the clipboard sites, so the `LoginItemSettings` re-export through `packages/shared/types.ts` was a non-event.

44.2.0 and 44.3.0 declare no breaking changes. Four behavior changes landed in them without a breaking-change note and were each read against the code; none reaches it. The one that looked closest was `execCommand('paste')` now needing transient activation in the calling frame, since `enableDeprecatedPaste` exists here for Google Docs — it falls back to a `clipboard-read` check, which `packages/app/account.ts` grants. The `setDisplayMediaRequestHandler` crash fix does not cover a crash Meru could hit either: the null dereference sits behind a `web_contents_id` that a screen or window grant never populates.

## The verification-code bug

`copyText` logged a failed write and resolved anyway, so awaiting it ordered the clipboard write ahead of marking the email read and deleting it without gating them on it. A failed write still deleted the email carrying the code — the outcome the await was there to prevent. It answers whether the write landed now. The same handler also read the view through the getter that throws once `destroy()` has cleared it. Not for the window the await opens, which is only as long as a clipboard write — `destroy()` leaves a notification that is already on screen alone, and this handler is that notification's click and action callback, so removing an account while a verification-code notification is showing and then clicking it runs the whole thing against a destroyed `Gmail`. The call is not awaited either way, so the throw would have surfaced as an unhandled rejection.

## `bun test --isolate`

`mock.module` replaces a module for the whole process rather than for the file that calls it, so two test files mocking `electron` for different reasons overwrite each other. The flag has to be typed — bun's `[test]` bunfig section has no key for it — so it goes on the CI step, the two package test scripts and the instruction file. A bare `bun test` still shares one registry.

## Verified

`fmt:check`, `lint`, `types`, 1056 unit tests and the end-to-end suite pass on all three platforms in CI, and locally on Linux under `xvfb` (45 passed, 4 skipped) against a packaged build. The e2e job builds the app before launching it, so macOS and Windows both compile and start on Electron 44. Playwright 1.62.1 drives it with no bump. Three cold-launch perf samples put total working set and the main JS heap flat against 44.1.1, with no bundle moving.

electron-builder is main's `^26.16.1` now rather than the `26.15.6` the gate was written against, so the premise it rests on was re-read: `minimumSystemVersion` still reaches only `Info.plist` and a `pkg` requirement, `publish/updateInfoBuilder.js` never writes it into `latest-mac.yml`, and electron-updater 6.6.2 still compares the feed value against `os.release()`, the Darwin version.

## Not verified

The three Gmail DOM-coupled features — the undo-send toast, the promotional banner hiding, the verification-code extractor — need a real signed-in session and are the main open risk; no Mac or Windows machine was available for hands-on work, so nothing exercised them.

CI builds unsigned, so the macOS signing and entitlements path still has not run anywhere, and the ANGLE static linking changes the file set it walks. `LSMinimumSystemVersion` has not been read back off a built `Info.plist`. The Monterey refusal itself is untestable here — the runners are far newer — so neither the plist declaration nor the updater gate has been seen to refuse anything on a real macOS 12 machine. The full-screen close workaround is not covered by the suite, so it is unverified under Chromium 152. The 1Password manual pass was not repeated, and no updater run from 3.61 was performed on any platform.


timche:
@coderabbitai review

PR #1059, opened 2026-09-04
Take conf 15 so migrations run after the defaults land

timche:
## Summary
Stacked on #1058. conf 14 throws away every migration's writes on the first launch after an upgrade that adds a default key, and the ladder only sticks on the second launch. conf 15 writes the defaults to disk before the ladder runs, which fixes that and the empty-store trap #1058 guards against. No ladder code changes — both existing patches carry over, and the tests move with the new ordering.

Review #1058 first; this needs its `config-migrations.ts`.

## Problem
conf 14.0.0's constructor snapshots the config file, computes `defaults + snapshot` in memory, runs the migration ladder, and then compares the two snapshots and writes the merged one back if they differ (`node_modules/conf/dist/source/index.js:96-110`). Neither snapshot reflects the migration, so any release that adds a default key overwrites the migration's writes with the pre-migration state — `__internal__.migrations.version` included. The file then holds every default, so the ladder runs again on the next launch and sticks.

Measured against Meru's own defaults and a 3.59.0-shaped config: launch 1 leaves `workspaceApps.launcherApps` as `["calendar", "notebooklm"]` with `__internal__` at 3.59.0, launch 2 gives `["calendar", "notebook"]` and 3.60.0.

It is long-standing — 3.58.0's `googleApps.*` → `workspaceApps.*` rename had the same one-launch delay, and a rename nothing reads is survivable. 3.60.0 is the first release where the stale state is *invalid* rather than merely old: `notebooklm` is not a key in `workspaceApps`, and `workspaceApps[app].label`, `.url` and `.singleInstance` are read unguarded (`packages/shared/google.ts:4`, `packages/renderer/components/vertical-tabs.tsx:156`, `packages/app/workspace-app.ts:404`). A user with a NotebookLM saved tab, launcher entry or bookmark therefore gets one launch of `undefined` app lookups.

## Solution
- `electron-store` 10.1.0 → 11.0.2, which is `conf` ^14 → ^15 and `type-fest` ^4 → ^5 and nothing else: electron-store's own `index.js` is byte-identical between the two versions. This repo already depends on `type-fest` ^5.2.0 directly, so that half drops a duplicate copy.
- conf 15's `#runMigrations` writes `defaults + file` to disk before `_migrate`, so a migration reads the current defaults rather than an empty store. It also deep-clones the rollback backup with `structuredClone` instead of a shallow spread, which matters here — six migrations mutate `accounts[]` in place.
- Both patches carry over unchanged in substance. conf 15 ships the same loose `get`/`set`/`has`/`delete`/`onDidChange` overloads the conf 14 patch already strips, plus one on the new `appendToArray`; leaving them in makes `set(key: string, value: unknown)` and the dot-notation `delete` accept any string, which turns 17 of the ladder's `@ts-expect-error` directives into unused-directive errors and silently drops the compile-time check on every removed key. The electron-store patch is the same single line letting `projectVersion` through.
- The guards in #1058 are not made redundant by this. Measured with the ladder as it stands on main — no guards — against conf 15: a fresh profile launches and an upgrade migrates on the first launch, but a stored config whose account has no `workspaceApps` still throws on `account.workspaceApps.savedTabs`. conf merges the defaults at the top level only, so a stored `accounts` keeps whatever shape it was written with and no default reaches inside it. The bump fixes the ordering; the guards cover the shape.
- The empty-store tests now run with no defaults at all. Under conf 15 a fresh profile is no longer an empty store, so testing through the defaults would have quietly stopped exercising the guards; passing none keeps them under test whatever conf does with ordering. Reverting a guard still fails two tests.
- The test that pinned the two-launch behavior now asserts the opposite, and the rename and promotion cases drop their throwaway first launch.

Considered and rejected: a two-phase construction in `config.ts` — seed the defaults with one `Store`, then construct again with the migrations — which is what conf 15 does internally. It avoids the dependency bump, but it is a hand-rolled version of an upstream fix, it would have to stay until conf is upgraded anyway, and the bump turned out to cost no application code at all.

## Try it
Nothing to open. `bun test packages/app/lib/config-migrations.test.ts` — 26 tests. The behavior change is visible as the diff to the "migrates on the first launch, not the second" case, which fails on conf 14 and passes on 15.

## Not verified
- Not run in a packaged app on any platform. This branch carries the 3.60.0-beta.1 version commit, so CI's `e2e` job launches it on a fresh profile on all three platforms, but no upgrade-over-3.59.0 test exists in CI and none was run by hand — the launch-1 behavior this fixes is the upgrade path specifically.
- conf 15 changes `clear()` twice — null-valued defaults are now restored rather than dropped, and the `store` setter re-injects `__internal__` from the existing file, so the migration version survives a reset where it used to be wiped. `packages/app/index.ts:65` is the only caller and it calls `app.relaunch()` immediately after, so neither should be observable; the Reset app path was not exercised.
- `config.clear()` (`packages/app/index.ts:65`, the Reset app path) changes twice under conf 15: the `store` setter now re-injects `__internal__` from the existing file, so a reset no longer re-runs the ladder on the next launch, and `isExist` dropped its null check, so null-valued defaults (`licenseKey`, `doNotDisturb.*`) are written rather than omitted. Both verified. Both should be invisible — `app.relaunch()` follows immediately, and conf 14 merged the defaults back on the next launch anyway — but the Reset app path was not exercised by hand.
- conf 15 changes the file watcher (directory watching on macOS, a 1s rather than 5s debounce on Linux) and adds `encryptionAlgorithm`, schema-violation handling under `clearInvalidConfig`, and prototype-pollution guards in `_set`. Meru passes none of `watch`, `encryptionKey`, `schema` or `clearInvalidConfig`, so all of that is inert here and was not tested.
- A fresh install now runs the whole ladder against the defaults rather than an empty store. None of the eighteen migrations rewrites a value the current defaults carry, so this is a no-op today; a future migration that did would change new-install behavior.


timche:
Deferred to the release after 3.60.0, where it can get the testing a store upgrade deserves rather than riding an overdue release.

The bug it fixes is real and unfixed: conf 14 discards every migration's writes on the first launch after an upgrade that adds a default key, so the ladder only takes effect on the second launch. For 3.60.0 that means one launch with `notebooklm` still in saved tabs, launcher entries and zoom factors, where `workspaceApps[app].label`, `.url` and `.singleInstance` are read unguarded. It self-corrects on the next launch. Recorded in the release risk review.

The branch stays for the next cycle. It carries the electron-store 11.0.2 bump, both ported patches, the migration ladder extracted to `packages/app/lib/config-migrations.ts`, and the test file — measured at zero application code beyond comments. `config-migrations-coverage` holds the extraction and the tests without the bump, if those want to land separately first.

PR #1074, opened 2026-09-11
Sign the Windows build with Azure Artifact Signing

timche:
Windows users can't add Meru to 1Password because the Windows build is unsigned, and 1Password's Add Browser only accepts a code-signed executable. This signs the Windows installers with Azure Artifact Signing.

- `win.azureSignOptions` in `package.json`, authenticated in the release workflow's Windows step with `AZURE_TENANT_ID`, `AZURE_CLIENT_ID` and `AZURE_CLIENT_SECRET`, which are set on the repo.
- The e2e build passes `signExecutable=false`, since the Azure config makes every Windows build sign and CI has no credentials there.
- The 1Password setup dialog now points Windows users at `%LOCALAPPDATA%\Programs\Meru` instead of `C:\Program Files`, where a per-user install never is.
- `verifyUpdateCodeSignature` stays `false` so existing unsigned installs update onto the first signed release.

The manual "Signed Windows build" workflow landed on main in #1078 and was run against this branch: https://github.com/zoidsh/meru/actions/runs/34860644602. All seven executables signed, and the installer's signature reads `CN=Zoid Ltd, O=Zoid Ltd, L=London, C=GB` with a Microsoft RFC 3161 timestamp, chained to the Microsoft Identity Verification root.

Draft until the artifact from that run has been installed on Windows and 1Password's Add Browser accepts it. arm64 signing runs the same path and is first exercised by the release.

Passkey sign-in on Windows is a separate issue: nothing in Meru blocks WebAuthn there and signing doesn't affect it.


timche:
Verified on Windows from the signed build artifact: installed, 1Password's Add Browser accepted Meru, and the extension works.

